PR #58, opened 2015-01-06
Indexing Event ID notes and linking to them like facets (DHH-597)

escowles:
Fixes https://lib-jira.ucsd.edu:8443/browse/DHH-597


mcritchlow:
This looks great. While we wait on confirmation for the facet display, would you mind adding a test that asserts an example record with an Event ID has this link displayed, and that one without an Event ID doesn't? It would be helpful as a reference during implementation of the next version of the DC app as well, noting that we have this use case.

PR #60, opened 2015-01-09
Fixing the direct object link when the referrer is not a search. DHH-582

VivianChu:


app/controllers/dams_objects_controller.rb
```diff
@@ -16,17 +16,21 @@ def show
...
     end
 
-	search_results = request.env["HTTP_REFERER"]	
-    session[:search_results] = search_results if (!search_results.nil? && search_results.include?("search"))
-      
-    # import solr config from catalog_controller and setup next/prev docs
-    @blacklight_config = CatalogController.blacklight_config
-    setup_next_and_previous_documents
-
+	  search_results = request.env["HTTP_REFERER"]
```
escowles: This looks good, except that this would call setup_next_and_previous_documents twice in the typical case where the link has a counter param and the referrer is /search.  So this should probably be an else if block.

As long as you're doing that, please clean up the indenting to use only spaces.

VivianChu:
@escowles The code has been updated. Thanks

PR #63, opened 2015-01-23
Removed repeated/duplicated names from the search result page

lsitu:
See https://lib-jira.ucsd.edu:8443/browse/DHH-590.

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -30,6 +30,16 @@ module CatalogHelper
       url = dams_object_path(doc, :counter => opts[:counter] )
     end
 
+    # Remove duplicated names
+    names = doc['name_tesim']
+    if (!names.blank?)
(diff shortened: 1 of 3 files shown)


app/helpers/catalog_helper.rb
```diff
@@ -30,6 +30,16 @@ def link_to_document(doc, opts={:label => nil, :counter => nil, :results_view =>
       url = dams_object_path(doc, :counter => opts[:counter] )
     end
 
+    # Remove duplicated names
+    names = doc['name_tesim']
```
mcritchlow: Consider Array#uniq method http://www.ruby-doc.org/core-2.2.0/Array.html#method-i-uniq

lsitu: Done! See commit https://github.com/ucsdlib/damspas/commit/c37da91d4a44c15922141ed52ee075a51962a3b9

PR #64, opened 2015-01-24
Adding utility to show ruby version

escowles:


mcritchlow:
What's the use case for this?

Does it make sense to render the version in JSON instead of HTML?

escowles:
The use case so far is developers/sysadmins wanting to know what version of ruby an app is actually running.  I'm not opposed to making a JSON version of this, but so far humans are the only consumer.  Maybe Nagios/etc. will be the next consumer, though.

mcritchlow:
Gotcha. In that case this definitely is a nice addition as-is. I suppose
another alternative if we have a need for non-human consumers, so no
parsing is needed, is to simply render the version out as the only content
on the page. Example of this I've seen in the wild for setting up a local
environment: https://github.com/thoughtbot/laptop/blob/master/mac#L159

PR #73, opened 2015-02-23
Added support for N-Triples and Turtle RDF formats.

lsitu:
Hi Esme,
As we discussed during our standup meeting last Thursday, I assume that damspas will be retrieve the N-Triples and Turtle RDF formats from damsrepo for now: https://lib-jira.ucsd.edu:8443/browse/DHH-626

### DIFF
--- a/app/controllers/dams_resource_controller.rb
+++ b/app/controllers/dams_resource_controller.rb
@@ -68,12 +68,16 @@ class DamsResourceController < ApplicationController
         format.html # show.html.erb
         format.json { render json: @document }
         format.rdf { render xml: @rdfxml }
+        format.nt { rdf_nt }
+        format.ttl { rdf_ttl }
       end
     elsif !@document.nil? && @document['discover_access_group_ssim'].include?("public")
       respond_to do |format|
         format.html { render :metadata }
         format.json { render json: @document }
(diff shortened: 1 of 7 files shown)


app/controllers/dams_resource_controller.rb
```diff
@@ -103,6 +103,18 @@ def rdf
     params[:xsl] = "normalize.xsl"
     data = get_html_data params, nil
     render :xml => data
+  end
+  def rdf_nt
```
escowles: Can you also add support for content-negotiation:

https://github.com/ucsdlib/damspas/blob/feature/rdf_formats/app/controllers/dams_resource_controller.rb#L67

Ideally, we should be able to either add .nt/.ttl extensions to the URL, or use an Accept header to trigger the different formats.

mcritchlow: +1

lsitu: Hi Matt and Esme, 
The feature for content negotiation with adding .nt/.ttl extensions to the URL is implemented. See https://github.com/ucsdlib/damspas/commit/bbca23a86ecb927f70f6cf9bef4062619179d03d.

PR #78, opened 2015-03-04
Allow for download of more than one master file. DHH-559

VivianChu:


app/views/dams_objects/_data_viewer.html.erb
```diff
@@ -11,6 +11,9 @@
 	%>
 
 	<div class="form-actions dams-download-button">
+        <% if (defined?(sourcefilePath) && sourcefilePath != filePath)%>
+            <a class="btn <%=btnColor%> pull-right-space" href="<%= sourcefilePath %>"><i class="icon-download-alt icon-white"></i> Download TGZ file</a>
```
escowles: The use case right now is a TGZ file, but we could have second master TIFFs or other formats that follow this pattern.  So maybe it makes more sense to get the extension from the filename, or get a label from the metadata instead of hard-coding TGZ.

VivianChu:
@escowles The label has been updated.  Thanks for the suggestion.

PR #80, opened 2015-03-07
Hiding counter param in the object view url. DHH-635

VivianChu:


escowles:
Closing this PR in favor of #81 (which adds feature tests that work without pre-loaded sample records, and demonstrates regression with page showing up on direct-linked pages)

PR #90, opened 2015-03-23
Adding Coveralls gem. DHH-633

VivianChu:


.coveralls.yml
```diff
@@ -0,0 +1 @@
+repo_token: YB4iUiX9WHTj4g24kZnb6o5IcUNg3R9jh
```
mcritchlow: I don't think we want this publicly available. Can we put in private_config, or otherwise?

VivianChu:
The .coveralls.yml has been moved to private_config.

PR #92, opened 2015-03-25
Fixed the status 404 error caused by referer controler handling with local access.

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-653

### DIFF
--- a/lib/dams/controller_helper.rb
+++ b/lib/dams/controller_helper.rb
@@ -675,8 +675,12 @@ module Dams
     def referrer_controller( request )
       uri = URI(request.referrer || "")
       if uri.host == request.host
-        ref = Rails.application.routes.recognize_path(uri.path.gsub(/^\/dc/,""))
-        ref[:controller]
+        begin
+          ref = Rails.application.routes.recognize_path(uri.path.gsub(/^\/dc/,""))
+          ref[:controller]
+        rescue Exception => e
+          logger.warn "Referer controller error: #{e}"
+        end
       end
     end
   end
--- a/spec/controllers/dams_objects_controller_spec.rb
+++ b/spec/controllers/dams_objects_controller_spec.rb
@@ -63,5 +63,12 @@ describe DamsObjectsController do
       expect(response.status).to eq( 200 )
       expect(session[:search_results]).to be_nil
     end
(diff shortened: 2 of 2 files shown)


lib/dams/controller_helper.rb
```diff
@@ -675,8 +675,12 @@ def audit( id = "unknown" )
     def referrer_controller( request )
       uri = URI(request.referrer || "")
       if uri.host == request.host
```
mcritchlow: Would it be possible to remove the begin/rescue block, and instead update the conditional check to include something like the following to ensure that the referrer URL is under the /dc/ application path?

`
if( uri.host == request.host && uri.path.start_with? '/dc' )
`

lsitu: I tried it earlier but some tests failed. I think this is because /dc is environment specific and our test environment don't have the /dc url pattern, so it failed to retrieve the controller from the referrer url.

mcritchlow: I see, that makes sense in the test context.

My concern is the begin/rescue block, while functional, doesn't really tell
us what purpose it serves by simply reading the code. Would something like
the Rails.root work instead of the hardcoded '/dc'?

lsitu: It looks like Rails.root is the path to the rails app but not the application root, and we can think about using RAILS_RELATIVE_URL_ROOT but I am seeing it ambiguous in test environment with a root like /. Maybe we can think about fixing the failed tests, or testing those failed tests in other ways?

mcritchlow: Yes, I think that would be a way to look into this. It would be nice if the code and tests could use whatever the URL base is for the application. Given that we need to get this in the release today for Matt P to continue his work, I'll merge this as-is, and setup a ticket to see if we can remove these inconsistencies in the next Sprint. Sound good?

lsitu: Yes, it sounds good. We can try to see whether other ways will work since Rails.application.routes.recognize_path() do throw the routing exception and the begin rescue block does seem to be correct here. But if we think about referrer_controller method here is just try to parse the referrer URL and return the controller if there is a matched otherwise return nil sounds easier to understand.

PR #94, opened 2015-04-14
Updated the date display label for the issue date and the collected date.

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DCMKEY-493

### DIFF
--- a/app/views/shared/fields/_date.html.erb
+++ b/app/views/shared/fields/_date.html.erb
@@ -20,7 +20,14 @@
 %>
 	     <% if date['type']!=nil && date['type']!=''&& date['value']!=nil &&date['value']!=''%>
 	     
-		    <% datetype= date['type'].capitalize+' Date' %>
+		    <%
+		      date_labels = t('date_labels')
+		      type = date['type']
+		      datetype = date_labels[type.to_sym]
+		      if (datetype.blank?)
+		        datetype = type.capitalize + ' Date'
+		      end
+		    %>
 		    <% htmlOpen = htmlOpenTemp %>
 		    <% htmlOpen %= datetype %>
 			<% concat htmlOpen.html_safe %>
--- a/app/views/shared/fields/_date.html.erb
+++ b/app/views/shared/fields/_date.html.erb
@@ -23,10 +23,7 @@
 		    <%
 		      date_labels = t('date_labels')
 		      type = date['type']
-		      datetype = date_labels[type.to_sym]
-		      if (datetype.blank?)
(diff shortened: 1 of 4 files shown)


spec/features/dams_object_spec.rb
```diff
@@ -505,4 +505,9 @@
     expect(page).to have_selector('button#node-btn-1:first',:text=>'Parameters')
     page.should_not have_css("button#node-btn-1", :count => 2)
   end
+  it "should not see label Creation Date and Issued Date" do
```
mcritchlow: What's the intended outcome of the test here? The text says "should not see Creation Date and Issued Date" but the #expect assertions are showing the opposite. Should the text change to "should see", or should the test assertions change?

lsitu: Good catch. I've corrected it to "should display". See https://github.com/ucsdlib/damspas/commit/34cefa16a526f05a42aaf9e082abaf223177a891.

config/locales/en.yml
```diff
@@ -22,3 +22,6 @@
 en:
   hello: "Hello world"
   sensitive_message: "Culturally sensitive content: This is an image of a person or persons now deceased. In some Aboriginal Communities, hearing names or seeing images of deceased persons may cause sadness or distress, particularly to the relatives of these people.  Would you like to view its content?"
+  date_labels: {
```
mcritchlow: This looks great. However, can we simply add all 5 type/label combinations in here? Then we have a single place to point our metadata staff to, should they want changes made.

app/views/shared/fields/_date.html.erb
```diff
@@ -20,7 +20,14 @@
 %>
 	     <% if date['type']!=nil && date['type']!=''&& date['value']!=nil &&date['value']!=''%>
 	     
-		    <% datetype= date['type'].capitalize+' Date' %>
+		    <%
+		      date_labels = t('date_labels')
+		      type = date['type']
+		      datetype = date_labels[type.to_sym]
+		      if (datetype.blank?)
```
mcritchlow: You can replace the blank conditional with #fetch if you'd like.

config/locales/en.yml
lsitu: Hi Matt,
It's done. See https://github.com/ucsdlib/damspas/commit/1267800c2edf2879ddea2f8055ff6ca0b7d13067.

mcritchlow:
:+1:  Thanks Longshou. This looks great!

PR #95, opened 2015-04-17
Fixed the missing general notes issue. 

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DM-97

### DIFF
--- a/app/views/shared/fields/_note.html.erb
+++ b/app/views/shared/fields/_note.html.erb
@@ -18,7 +18,6 @@
 
   # parse notes
   notes = Hash.new
-  notes.compare_by_identity
   
   # Temporary fix the bug that note field not displaying at component level. 
   # The solr index will need to be updated to make it consistent.
@@ -34,6 +33,7 @@
       note_type = note['type']
       note_label = note['displayLabel']
       note_internal = note['internalOnly'] == 'true'
+      key = 'General Note'
       # add notes[displaylabel] if exist, if not note[type] else put as general note
       # add notes if type/label match specification
       if ( type == nil || type == note_type ) &&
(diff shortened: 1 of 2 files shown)


app/views/shared/fields/_note.html.erb
```diff
@@ -46,7 +46,7 @@
         elsif note_type != nil && !note_type.blank?
           notes[note_type.titleize] = note
         else
-          notes['General Note'] = note
+          notes['General Note'.to_s] = note
```
mcritchlow: I'm not following turning a String into a String again as the solution. Apologies if I'm missing something.

I believe you noted the problem is that if there are more than one instance of 'General Note' then all but one are being overwritten by this assignment.

Rather than use '=' assignment, is this an instance where using '<<' assignment would be correct? This way you would never write over a previously stored value, you're just appending values. In fact, shouldn't we be doing this is all 3 conditional scenarios: https://github.com/ucsdlib/damspas/blob/feature/note_missing/app/views/shared/fields/_note.html.erb#L44-L50

lsitu: I think the magic is in line https://github.com/ucsdlib/damspas/blob/feature/note_missing/app/views/shared/fields/_note.html.erb#L21 for Hash#compare_by_identity: notes.compare_by_identity, which is always use the reference id of the key value (string). So it's always inserting a new key => value pairs to the Hash. Then line https://github.com/ucsdlib/damspas/blob/feature/note_missing/app/views/shared/fields/_note.html.erb#L56 to group them together.
Yes, I think we can change the lines to << insert new notes to the array instead if you think it'll be better. What do you think?

mcritchlow: Thanks for explaining. I think that while creating a new String, and thus a new object ID, does work, it is quite cryptic.

Appending the values to each note array is going to be more intuitive for our team to read in the future if other changes are needed. It makes perfect sense in many cases that we end up with an array of General Notes, and potentially other notes as well.

lsitu: Hi Matt,
I've changed it to use the Array data structure to process all the note types, which looks more straight forward for understanding. See https://github.com/ucsdlib/damspas/commit/3986ccf0b6f3f21f1553573c95d05b9867f1f448.

mcritchlow:
This a great improvement/simplification :+1: Thanks @lsitu

PR #96, opened 2015-04-22
Fixing the download issue for CEO content in curator mode. DHH-654

VivianChu:


app/views/dams_objects/_complex_object_viewer.html.erb
```diff
@@ -33,7 +33,7 @@
         if source_file != nil
           download_file_path = download_path(ark,"_#{source_file}")
         elsif service_file != nil
-          download_file_path = download_path(ark,"_#{service_file}")
+          download_file_path = service_file_path
```
mcritchlow: Is download_path not working because of the naming convention change between objects and components? I'm wondering why service_file_path is proposed here vs fixing the call to download_path to use _#{i}_#{service_file}"

Or, put another way, why not do the same variable assignment for the source and display file as proposed for service files? 

I think we should be consistent about how the paths are being assigned, one way or the other.

VivianChu:
The download_file_path has been updated for consistency.

mcritchlow:
:+1: Thanks Vivian!

PR #99, opened 2015-05-11
Adding support for suppressDiscovery restriction

escowles:
Fixes https://lib-jira.ucsd.edu:8443/browse/DHH-670


mcritchlow:
@lsitu and @VivianChu How does this look to you? Any concerns?

PR #100, opened 2015-05-11
Add Collection info to the search results view. DHH-664

VivianChu:


app/controllers/catalog_controller.rb
```diff
@@ -144,10 +144,10 @@ def exclude_unwanted_records(solr_parameters,user_parameters)
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
 	#	UCSD custom added argument :hitsonly to display the values that has hit text.
+    config.add_index_field 'collection_name_tesim', :label => 'Collection:', :highlight => config.highlighting
```
mcritchlow: Hi Vivian, why was the field name changed from 'collection_1_name_tesim' to 'collection_name_tesim'?

VivianChu:
Hi Matt, the field collection_1_name_tesim was missing from the search results.  Only the field collection_name_tesim exists.
For example http://lib-hydratail-qa:8080/solr/blacklight/select?q=fish&wt=xml&debugQuery=true&qf=all_fields_tesim

mcritchlow:
Thanks Vivian, how strange. Perhaps this changed over time with our to_solr indexing strategy?

mcritchlow:
Following up, Arwen confirmed that they would like to have all collections displayed in the UI. Vivian, can you please:
- Create an additional test for an object that is part of multiple collections
- Attach a screenshot to the JIRA ticket to show Arwen what the UI would look like in that case?

Once we've confirmed the look and behavior, I'll merge this PR. Thanks!

VivianChu:
Hi Matt,  I added an additional test and attached the screen shot for that case.
Thanks

PR #104, opened 2015-05-19
Fixing a few tests

escowles:
Related to DAMSKEY-509


mcritchlow:
@escowles I'm seeing conflicts here that prevent GH from automatically merging. Can you take a look?

escowles:
I've rebased and it should merge cleanly now.

PR #107, opened 2015-05-21
Upgrade to Rails 4.2.1 DHH-663

VivianChu:


app/models/datastreams/dams_resource_datastream.rb
```diff
@@ -249,11 +249,11 @@ def insertNameFields (solr_doc, fieldName, objects)
         insert_unique_field_value solr_doc, "name", obj.name 
       end
     end
-    reload MadsPersonalNameInternal
-    reload MadsNameInternal
-    reload MadsConferenceNameInternal
-    reload MadsCorporateNameInternal
-    reload MadsFamilyNameInternal
+    #reload MadsPersonalNameInternal
```
escowles: Please go ahead and remove these lines instead of commenting them out.

config/environments/production.rb
```diff
@@ -10,7 +10,7 @@
   config.action_controller.perform_caching = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this)
-  config.serve_static_assets = false
+  config.serve_static_files = false
```
escowles: We need to make sure this change is also made in the private_config versions of the environment files.

VivianChu: @escowles The change has been made to the private_config versions of the environment files.

PR #109, opened 2015-05-27
Fixes DHH-678

VivianChu:


spec/features/dams_collections_spec.rb
```diff
@@ -64,3 +64,23 @@
...
+  end
+  after do
+    @provCollection.delete
+    @unit.delete
+  end 
+  scenario 'should see the collection result page with no issued date' do
+    visit catalog_index_path( {:q => "#{@provCollection.pid}"} )
+    expect(page).to have_selector('h3', :text => 'Heavy Metals in the Ocean Insect, Halobates')   
+    expect(page).to have_selector("ul.dams-search-results-fields:first li span", :text => '1961-1978')     
+  end
```
escowles: Please add an explicit have_no_content test to make sure the issue date isn't displaying (you may want to change the value so other instances of 2015 don't throw it off).

app/helpers/catalog_helper.rb
```diff
@@ -86,6 +86,7 @@ def date_list( document )
         end
       end
     end
+    dateVal = dateVal[0,dateVal.length()-2] if !dateVal.blank? && dateVal.end_with?(', ')
```
escowles: This looks overly complex too me, and would still not handle dateVal being nil.  How about:

```
dateVal = dateVal.gsub(/, $/,'') if dateVal
```

VivianChu:
@escowles - The code has been updated.  Thanks.

PR #116, opened 2015-06-30
Upgrading to rails 4.2.2 and adding coveralls badge. (DHH-696, DHH-695)

hweng:


db/schema.rb
```diff
@@ -13,54 +13,54 @@
 
 ActiveRecord::Schema.define(version: 20140501112514) do
 
-  create_table "audits", force: true do |t|
-    t.string   "user"
-    t.string   "action"
-    t.string   "classname"
-    t.string   "object"
+  create_table "audits", force: :cascade do |t|
```
mcritchlow: were these force :cascade changes part of the Rails 4.2.2 upgrade?

hweng:
Yes, that was automatically changed with Rails 4.2.2 upgrade.

PR #125, opened 2015-08-03
Adding date facet, fixing date sorting feature test (DHH-632)

escowles:


Gemfile.lock
```diff
@@ -417,6 +417,3 @@ DEPENDENCIES
   uglifier (~> 2.5.3)
   unicode (= 0.4.4)
   unicorn (~> 4.8.3)
-
-BUNDLED WITH
```
mcritchlow: Are you removing this because of the issues noted here? http://bundler.io/blog/2015/06/24/version-1-10-released.html

escowles: No, I just ran bundle install as usual and it removed it.  I just checked and I had an old version of bundler (1.8.x), so I updated to 1.10.6 and it's adding the bundler version to the end of the file.  I'll push another commit that adds that back.

PR #126, opened 2015-08-17
Escaping fulltext before sending to solr (DHH-724)

escowles:
Fixes https://lib-jira.ucsd.edu:8443/browse/DHH-724
Also fixing deprecations in controller tests

### DIFF
--- a/lib/dams/dams_object.rb
+++ b/lib/dams/dams_object.rb
@@ -382,15 +382,23 @@ module Dams
           end
 	    }
 	  end
+
+      # extract fulltext from a content file
       def fulltext( object, file, cid )
         mime_type = file.mimeType.first.to_s.gsub(/;.*/,"")
         if ["application/pdf", "text/html", "text/plain"].include?(mime_type)
           fid = (cid != nil) ? "_#{cid}_#{file.id}" : "_" + file.id
-          fid = "fulltext#{fid}" if mime_type == "application/pdf"
+          fid = extracted_text(fid) if mime_type == "application/pdf"
           fulltext = @parent_obj.datastreams[fid]
           escape_text(fulltext.content) unless fulltext.nil?
         end
       end
+
+      # file_id of text extracted from a content file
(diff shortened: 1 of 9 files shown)


lib/dams/dams_object.rb
```diff
@@ -391,6 +382,23 @@ def insertFileFields( solr_doc, cid, files )
           end
 	    }
 	  end
+      def fulltext( object, file, cid )
+        mime_type = file.mimeType.first.to_s.gsub(/;.*/,"")
+        if ["application/pdf", "text/html", "text/plain"].include?(mime_type)
+          fid = (cid != nil) ? "_#{cid}_#{file.id}" : "_" + file.id
+          fid = "fulltext#{fid}" if mime_type == "application/pdf"
```
mcritchlow: why is the addition mime_type conditional here only relevant for PDFs?

escowles: HTML and plain text can simply be retrieved as-is -- PDFs need to have the fulltext prefix to trigger the DAMS REST API text extraction service with Tika.

mcritchlow: Thanks, makes perfect sense. Do you think it would be easier to come back and refactor/migrate this in the future if the fulltext extraction conditional assignment was a method that described the requirement? Something like:
```
add_fulltext_service_trigger(fid) if mime_type == "application/pdf"
.....
# Add a prefix to PDF file to trigger repository text extraction service
def add_fulltext_service_trigger(fid)
  "fulltext#{fid}"
end
```

Or, do you think we're moving far enough away from this via Sufia/F4 that it's not worth the hassle and refactoring/migrating won't be an issue?

escowles: I think the pattern is going to be pretty similar with Sufia/CurationConcerns: text/html can be used directly, but PDFs will need to have text extracted.  It'll be saved to a separate file by a background job, but it will look pretty much the same from the indexer's point-of-view.

PR #132, opened 2015-09-04
Updating most gems to most recent versions

escowles:
Not updating core Hydra gems and some older versions they require


mcritchlow:
:+1: thanks @escowles

PR #134, opened 2015-09-08
Component metadata refactor

escowles:
* Adding component parameter to primary metadata partial
* Replacing most of component metadata partial with call to primary metadata partial

Fixes https://lib-jira.ucsd.edu:8443/browse/DHH-726

### DIFF
--- a/app/views/shared/fields/_show_raw.erb
+++ b/app/views/shared/fields/_show_raw.erb
@@ -1,7 +1,7 @@
 <%
   component = defined?(componentIndex) ? {:componentIndex => componentIndex} : { }
 %>
-<% collectionsView = defined?(greetingsFromCollectionsView) ? true : false %>
+<% collectionsView = defined?(greetingsFromCollectionsView) %>
 <% notesToExclude = ['performers','extent','abstract','table of contents','identifier',
         'existence and location of originals','general physical description', 'dimensions',
         'digital origin', 'physical description'] %>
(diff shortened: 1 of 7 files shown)


app/views/shared/fields/_show_raw.erb
```diff
@@ -5,66 +8,71 @@
...
+<%= render :partial => 'shared/fields/note', :locals => {:type => 'abstract'}.merge(component) %>
+<%= render :partial => 'shared/fields/scope_of_content', :locals => component %>
+<%= render :partial => 'shared/fields/cartographic', :locals => component %>
+<%= render :partial => 'shared/fields/note', :locals => {:type => 'table of contents'}.merge(component) %>
 <%= render :partial => 'shared/fields/note', :locals => {:exclude_type => notesToExclude,
-        :exclude_label => ['digital object made available by'] } %>
-<%= render :partial => 'shared/fields/resource_type' %>
-<%= render :partial => 'shared/fields/language' %>
+        :exclude_label => ['digital object made available by'] }.merge(component) %>
+<% unless defined?(componentIndex) %>
```
mcritchlow: for consistency, can the resource_type partial also use the new Hash#merge approach rather than `unless defined?(componentIndex)`?

app/views/shared/fields/_show_raw.erb
```diff
@@ -5,66 +8,71 @@
...
+  <%= render :partial => 'shared/fields/license', :locals => component %>
+  <%= render :partial => 'shared/fields/statute', :locals => component %>
+  <%= render :partial => 'shared/fields/other_rights', :locals => component %>
 <% end %>
 
-<%= render :partial => 'shared/fields/rights_holder' %>
-<%= render :partial => 'shared/fields/unit' %>
+<%= render :partial => 'shared/fields/rights_holder', :locals => component %>
+<% unless defined?(componentIndex) %>
+  <%= render :partial => 'shared/fields/unit' %>
```
mcritchlow: for consistency, can the unit partial also use the new Hash#merge approach rather than `unless defined?(componentIndex)`?

escowles:
@mcritchlow, I've pushed a second commit to use the same locals approach, and it turned out that both the unit and resource_type partials could be greatly simplified because they were never used at the component level.

mcritchlow:
:+1: thanks @escowles!

PR #137, opened 2015-09-17
Adding creator to the name order.(DHH-728)

hweng:


spec/features/dams_collections_spec.rb
```diff
@@ -107,11 +107,13 @@
   end
   scenario 'should see the names in order' do
     visit dams_collection_path("#{@provCollection.pid}")
+    require 'pry'; binding.pry
```
escowles: Please remove the call to pry.

PR #141, opened 2015-09-25
Added "view file" button to object view (DHH-714)

dt-ucsd:


escowles:
@dt-ucsd: Can you add or update a test to demonstrate this update?  And please rename your branch to start with feature/ so Bamboo will build it automatically.

PR #145, opened 2015-10-14
Reload record when failed to fix the potential cause of missing solr field. DHH-605

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-605

### DIFF
--- a/lib/dams/dams_helper.rb
+++ b/lib/dams/dams_helper.rb
@@ -825,7 +825,7 @@ module Dams
     end
 
   # helper method to load external class objects
-  def loadRdfObjects (object, className)
+  def loadRdfObjects (object, className, reload: true)
     objects = []
     if !object.first.nil?
       object.each do |o|
@@ -838,8 +838,9 @@ module Dams
           begin
             o = className.find(o_pid)
             objects << o
-          rescue
-            logger.warn "Error loading  #{className.to_s}: #{o_pid}"
+          rescue => e
+             logger.warn "Error loading #{o_pid} in class #{className.to_s}: #{e}. Reloading..."
+             loadRdfObjects object, className, false if reload
           end
         end
       end


lib/dams/dams_helper.rb
```diff
@@ -838,8 +838,13 @@ def loadRdfObjects (object, className)
           begin
             o = className.find(o_pid)
             objects << o
-          rescue
+          rescue Exception => ex
             logger.warn "Error loading  #{className.to_s}: #{o_pid}"
+            # Try to reload the record when loading failed
+            if reload
+                puts "Error: #{ex}. \n Reloading #{o_pid} in class #{className.to_s}."
```
mcritchlow: :+1: to this approach! A couple of thoughts.

In this rescue block, in the situation where reload is true (i.e. trying a second time), then there will be two log entries which might be a bit confusing.

What if we changed the parameter from reload to something like:

`def loadRdfObjects(object, classname, tries: 2)`

then the rescue block could take advantage of the ruby `retry` keyword (see http://docs.ruby-lang.org/en/2.2.0/syntax/exceptions_rdoc.html for more info) and do something like:

```
rescue => e
  tries -= 1
  logger.warn "Error: #{e}. loading #{o_pid} in class #{className.to_s}. #{tries} tries left."
  retry if tries > 0
end
```

also this scenario should use `logger.warn`, or possibly even `logger.error`, instead of `puts ...` as this would be a problem that we would want to elevate as it would mean an existing service is down or not responding.

lsitu:
@mcritchlow It looks more cleaners and I've replaced mime with you suggested codes. Thanks. 
I just want to ensure that reload effort is logged by the command line solrizer by using "puts" instead of logger.warn since logging may depend on how the Solrizer is started. I've changed to use logger.warn.

mcritchlow:
:+1: looks good. i'll wait until the bamboo build finishes to make sure the test suites still passes with the change. so, merged and ready by tomorrow morning at the latest.

lsitu:
@mcritchlow, I've changed it to reload the whole method instead of using retry to load the last failed one to see how it goes: https://github.com/ucsdlib/damspas/commit/31ffa3a8e4b4a29ca7fe1f8e37c83ce4ee771678 .

mcritchlow:
@lsitu thanks. per our discussion, just to note here, while using `retry` would be ideal you've switched to rerun the entire method to be as safe as possible while we're trying to isolate the problem. assuming we do, in the future we may want to refactor to use `retry` rather than re-running the entire method (and inner loop for loading all linked objects).

PR #146, opened 2015-10-15
Fixing PDF view/download file tests.(DHH-750)

hweng:


spec/features/dams_object_spec.rb
```diff
@@ -427,6 +421,28 @@
     end
   end
 
+  describe "Curator PDF Viewer" do
```
mcritchlow: @hweng this is excellent, thanks for doing this so quickly. the only issue i notice is that I don't believe this feature is limited to curators, based on discussion in https://lib-jira.ucsd.edu:8443/browse/DHH-714 was there something you read in the ticket that called out curators specifically? if so, we should confirm with Gabriela

hweng:
@mcritchlow, let me confirm with Gabriela, thanks!

mcritchlow:
@hweng looks good! will merge once bamboo finishes running the tests

mcritchlow:
bamboo is :smile: again. thanks @hweng !

PR #148, opened 2015-10-19
Adding support for decade facet multivalued indexing(DHH-733)

hweng:


mcritchlow:
looks good, waiting to see about bamboo build issues. can you also please squash your commits for this feature [1] into one when we confirm the bamboo issues? thanks!

1. https://github.com/ucsdlib/DeveloperGuides/blob/master/protocol/git.md#write-a-feature

hweng:
@mcritchlow All tests passed on bamboo and for me locally.  There are two commits because I had more ideas to handle additional use case scenarios after committed first one.  And I tired squash commits before pull request, but it seems there is another commit about fixing reload error in between my two commits. I don't think it is good to mix them together. What do you think?  Thanks!

mcritchlow:
Though this PR looks great :+1: , I want to hold on merging until the develop branch is green again. hopefully @lsitu will have time to look into that.

Regarding squashing the commits, let's see what is needed to get the develop branch green again and evaluate from there.

PR #149, opened 2015-10-22
Updated Solrizer to retry on error and skip indexing the record if still failed. DHH-605

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-605

### DIFF
--- a/bin/damsolrizer
+++ b/bin/damsolrizer
@@ -89,6 +89,8 @@ begin
   $stderr.print "Getting output from #{@destination} (priority > #{@priority})\n"
   
   @conn.subscribe(@destination, {:ack =>"client" })
+
+  failed_count = 0
   while true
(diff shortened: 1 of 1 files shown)


mcritchlow:
@lsitu :+1:  i'll merge. if you can please test this on QA asap to confirm the retry functionality will work and then fail gracefully as you've outlined in this change that would be great. i'd like to have confirmation this won't introduce any regression indexing issues prior to a release.

lsitu:
@mcritchlow I've run the same code on QA for sometime this morning and it's moving smoothly with errors logged. Now I restarted the Solrizer with the latest damspas code on QA, I see it works with Date parsing error logged. So I think the latest damspas deployment on QA should be good.

mcritchlow:
That's great, thanks @lsitu!

PR #150, opened 2015-10-29
 Fixing collection facets missing from advanced search form. (DHH-751)

hweng:


mcritchlow:
this is definitely an approach worth trying out. :+1: 

unfortunately code climate checking failed, not sure why.

i'm also seeing a single test failure here from bamboo http://lib-bamboo.ucsd.edu/browse/DHH-PRSPEC42-2/test

we can re-run the suite on the branch to confirm, but wanted to confirm whether you're seeing this on your end.

hweng:
@mcritchlow All tests passed on my local computer. I'm not sure why code climate checking failed either. I just re-run code climate analyzing to see what happens.

mcritchlow:
@hweng sounds good. looks like you kicked off a new run in bamboo too, so we'll see how those results end up and go from there.

mcritchlow:
@hweng unfortunately the test is still failing http://lib-bamboo.ucsd.edu/browse/DHH-PRSPEC42-3

somewhat related, it looks like the test is bypassing the UI and the new conditional code you've added. could the test, instead of going directly to the advanced search page, could we confirm the behavior change you've introduced by going to browse by collection first and following the navigation to Advanced Search?

So instead of:

`visit '/advanced?spellcheck.q='`

Something like (and there may very well be a better way to do this):

```
visit 'dc/collections'
advanced_search_link = find(:css, "a:contains('Advanced Search')")
advanced_search_link.click
# ... expect#have_selector assertions for Collection,Format,Repository
```

hweng:
@mcritchlow Your suggestion sounds good to me. I just re-run the test suit on my local and all tests passed, but it failed on bamboo which is strange.

hweng:
@mcritchlow, I've added test records for collection. Now the bamboo tests passed. 

And for visit link, I changed it to
```
visit '/collections'
click_link "Advanced Search"
expect(page)...........
```

mcritchlow:
:clap: looks great!

PR #157, opened 2015-12-03
Fixing related resource datastream model and solr index issues.(DHH-746)

hweng:


mcritchlow:
looking good!

could you please add at least one test that confirms multiple related resources will display for components? something like https://github.com/ucsdlib/damspas/blob/develop/spec/features/dams_object_spec.rb#L524 but with multiple resources?

hweng:
@mcritchlow The tests for multiple related resource in components has been added to dams_object_spec.rb.

PR #159, opened 2015-12-11
Add support for 1-m variants per Title. DHH-765

VivianChu:


lib/dams/mads_title.rb
```diff
@@ -197,9 +197,11 @@ def set_variant(type, val)
 	  end
 	  	  
 	  def get_variant(type)
```
mcritchlow: do you think we should rename this method given the change of scope? i know rubocop wouldn't love the get part either (i know this method far predates our use of rubocop), so maybe it could be `variants`?

VivianChu: Sure, I can rename it and update all the methods that use it.

PR #160, opened 2015-12-15
Added access control text to search results and object/collection views (DHH-753)

dt-ucsd:


spec/features/dams_collections_spec.rb
```diff
@@ -148,4 +155,28 @@
...
+
+  after(:all) do
+    @prov.delete
+    @part.delete
+  end
+
+  scenario 'should see access control information (curator view)' do
+    sign_in_developer
+    visit dams_collection_path @prov.pid
+    expect(page).to have_content('AccessCurator Only')
```
mcritchlow: Should this be: `Access: Curator Only`?

spec/features/dams_object_spec.rb
```diff
@@ -685,4 +685,26 @@
...
+    solr_index (@ctsObject.pid)
+  end
+  after do
+    @ctsObject.delete
+    @unit.delete
+  end
+  it 'with access control information on page (curator)' do
+    sign_in_developer
+    visit dams_object_path(@ctsObject.pid)
+    expect(page).to have_content('AccessCurator Only')
```
mcritchlow: Should this be `Access: Curator Only`?

spec/features/catalog_spec.rb
```diff
@@ -268,6 +268,40 @@
...
+    @obj1.delete
+    @obj2.delete
+    @obj3.delete
+    @copy.delete
+    @unit.delete
+    @sub1.delete
+    @sub2.delete
+  end
+
+  scenario 'with access control information (curator)' do
```
mcritchlow: Can we also add a test asserting that a public object result will not show the access information? That's another key requirement for this, that would be helpful to have a test for.

spec/features/dams_collections_spec.rb
dt-ucsd: No. It displays as "AccessCurator Only" in rspec because I am using <span> elements to style them in regular browsers.

mcritchlow: ah, ok. thanks for clarifying

spec/features/dams_object_spec.rb
dt-ucsd: No. It displays as "AccessCurator Only" in rspec because I am using [span] elements to style them in regular browsers.

mcritchlow:
@dt-ucsd thanks for adding those tests :+1: . I think the last todo's are:

* Update the UCSD Only text as Gabriela noted in the JIRA ticket today
* rebase/squash your commits to a single commit. you can leave the 'merge into develop' commit alone, but if you can squash the others that would be great. see https://github.com/ucsdlib/DeveloperGuides/blob/master/protocol/git.md#write-a-feature and in particular the rebase/squash link towards the bottom of that section.

PR #161, opened 2015-12-16
Add an API feature for a random public object/collection. DHH-763

VivianChu:


app/views/random/show.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= @docs %>
```
mcritchlow: @VivianChu so, this all is looking really good. 

I think you can simplify this somewhat, because we don't actually need to display this to end users in a show view. All that we would need to do would return this as a json response, since this request will be used either a) internally to get the URL and then use it to generate something like a Tweet to post or b) allow an external service to request the random URL. in either case, having the /random request return the URL as a JSON response should do the job. Apologies if that wasn't clear in the ticket.

VivianChu: I updated the code and moved the tests from feature to controller directory to test for json response. Thanks

PR #164, opened 2016-01-05
Simplified accessGroup conditional (DHH-773)

dt-ucsd:


lsitu:
:+1: All three failed tests passed. Thanks dt-ucsd.

PR #165, opened 2016-01-06
Add backward support for the current single variant string type. DHH-774

VivianChu:


lsitu:
:+1: Looks good, Vivian.

PR #166, opened 2016-01-12
Replace subcollections list with Solr query for the collection landing page. DHH-770

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-770

### DIFF
--- a/app/controllers/dams_resource_controller.rb
+++ b/app/controllers/dams_resource_controller.rb
@@ -29,8 +29,20 @@ class DamsResourceController < ApplicationController
 
     # get metadata from solr
     @document = get_single_doc_via_search(1, {:q => "id:#{params[:id]}"} )
+    if @document.nil?
+        raise ActionController::RoutingError.new('Not Found')
+    end
 
-    @rdfxml = @document['rdfxml_ssi'] if !@document.nil?
+    # generate facet collection list for collection page only
+    models = @document["active_fedora_model_ssi"]
+    if models.include?("DamsAssembledCollection") || models.include?("DamsProvenanceCollection") || models.include?("DamsProvenanceCollectionPart") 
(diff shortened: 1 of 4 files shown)


app/controllers/dams_resource_controller.rb
```diff
@@ -30,6 +30,17 @@ def show
     # get metadata from solr
     @document = get_single_doc_via_search(1, {:q => "id:#{params[:id]}"} )
 
+    # generate facet collection list for collection page only
+    if !@document.nil?
```
mcritchlow: @lsitu we have a number of instances in the show method where we're checking whether @document is nil or not. what are your thoughts on checking @document immediately after the `get_single_soc_via_search` method and returning if it is nil at that point? then we could remove all the other nil checks in the method.

lsitu: Done. I see it break the test in https://github.com/ucsdlib/damspas/blob/develop/spec/features/dams_object_spec.rb#L371-L374 with the 404 status so just keeping the conditional check unless we really want the 404 status. See commit https://github.com/ucsdlib/damspas/commit/afdfccad8fec5c37a1e1d88b99699bb108316518

mcritchlow: Interesting, I'm trying to recall if there's a reason we would have had that test assert a 403 instead of a 404. What do you think would be technically correct despite the existing test? It seems like 404 (not found) is what is actually happening here.

lsitu: Yeah, in this case the 403 seems confusing to me. For example, we may really want status 404 when a records was deleted and we want to do a double check to confirm the deletion.

mcritchlow: I agree. Given then, would you mind updating the behavior to throw a 404, and the test to expect a 404?

lsitu: I've updated it for status 404. See commit https://github.com/ucsdlib/damspas/commit/78d697388089371ed486a4bcc4500dc9fa47c9a2 .

PR #170, opened 2016-01-28
add solr and fedora cleanup for test suite

mcritchlow:
Currently, or test suite doesn't fully clean up after itself. Some documents remain in Solr, for example.

This update more intentionally clears out solr and fedora before and after the test suite runs.

Note that newer versions of active fedora have a [Cleaner](https://github.com/projecthydra-labs/activefedora-aggregation/blob/master/spec/spec_helper.rb#L7-L10) class built to do this, unfortunately we can't take advantage of that given the version of AF we're pinned to.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -45,6 +45,23 @@ RSpec.configure do |config|
   # the seed, which is printed after each run.
   #     --seed 1234
 #  config.order = "random"
+  config.before(:suite) do
+    cleanout_solr_and_fedora
+  end
+  config.after(:suite) do
+    cleanout_solr_and_fedora
+  end
+end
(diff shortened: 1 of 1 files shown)


mcritchlow:
@lsitu @hweng @VivianChu - while this won't address the bamboo failed tests we're seeing with branching stepping on top of each other, it is something Longshou and I noticed while looking into that problem. Feedback welcome, thanks!

lsitu:
:+1: It looks good to me.

VivianChu:
Looks good to me too :+1:

mcritchlow:
@lsitu @VivianChu thanks. if one of you can merge that would be great.

hweng:
:+1: Looks great.

PR #171, opened 2016-02-03
Added supplemental fulltext support for files like Word and Excel documents, .html, .txt and .csv etc. DAMSKEY-559

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DAMSKEY-559

### DIFF
--- a/spec/features/dams_object_spec.rb
+++ b/spec/features/dams_object_spec.rb
@@ -385,12 +385,13 @@ feature 'Visitor want to look at objects' do
   describe "viewing files" do
     before(:all) do
       @col = DamsAssembledCollection.create( titleValue: 'Test Collection', visibility: 'public' )
-      @o = DamsObject.create( titleValue: 'Image File Test', copyright_attributes: [ {status: 'Public domain'} ],
+      @o = DamsObject.create( titleValue: 'Object Files Test', copyright_attributes: [ {status: 'Public domain'} ],
                   assembledCollectionURI: [ @col.pid ], typeOfResource: 'image' )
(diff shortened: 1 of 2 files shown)


spec/features/dams_object_spec.rb
```diff
@@ -434,6 +435,10 @@
       visit catalog_index_path( { q: 'THsdtk', sort: 'title_ssi asc' } )
       expect(page).to have_selector('h3', :text => "Image File Test")
     end
+    it 'should index fulltext of complex object CSV file' do
+      visit catalog_index_path( { q: 'Testing CSV', sort: 'title_ssi asc' } )
+      expect(page).to have_content("Image File Test")
```
mcritchlow: The name of the test object with the title 'Image File Test' is a little misleading now since we're using that object to test more than image files and associated behavior. Given that it's used in the context of 'viewing files' should we alter the name to something more generic?  (https://github.com/ucsdlib/damspas/blob/feature/fulltext_index/spec/features/dams_object_spec.rb#L385-L442)

lsitu: I just changed the title "Image File Test to the more generic title "Object Files Test".

mcritchlow: :+1: that looks good. Will the following expect assertions need to change as a result?

- https://github.com/ucsdlib/damspas/blob/feature/fulltext_index/spec/features/dams_object_spec.rb#L432
- https://github.com/ucsdlib/damspas/blob/feature/fulltext_index/spec/features/dams_object_spec.rb#L436
- https://github.com/ucsdlib/damspas/blob/feature/fulltext_index/spec/features/dams_object_spec.rb#L440

lsitu: Yes. I realized it shortly after my first commit and I've corrected them.

mcritchlow:
:+1: thanks @lsitu!

PR #172, opened 2016-02-19
Adding SHARE API support.(DAMSKEY-521)

hweng:


spec/controllers/dams_collections_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      @provCollection.save!
+      solr_index (@provCollection.pid)   
+    end
+    after do
+      @provCollection.delete
+      @unit.delete
+    end
+
+    it "should return OSF json" do
+      get :osf_api, id: @provCollection.pid
```
mcritchlow: do you need to do a developer sign in here first?

hweng: @mcritchlow thanks for check it. I just added developer sign in but still got failed. The wired thing is that I ran many times of tests in share_api branch and all got passed, but after merge the develop branch into it the test got failed. So I've checking the recent changes in develop branch which might affect it.

mcritchlow:
:+1:

PR #173, opened 2016-02-24
Remove click-thru for embargoed content.(DHH-780)

hweng:


hweng:
move to https://github.com/ucsdlib/damspas/pull/174

PR #176, opened 2016-03-03
Feature/DHH-785

hweng:


mcritchlow:
the fix looks :+1: can you remove the rails commit for DHH-790? that's going to cause an issue with the other PR

app/views/shared/fields/_note.html.erb
```diff
@@ -56,22 +56,25 @@
 
   if !notes.empty?
     # if type is specified, bundle all matching notes under type header
-    if type != nil
+
+    if type != nil && type != 'related publications'
       htmlOpen = htmlOpenTemp
       htmlOpen %= type.to_s.capitalize
```
mcritchlow: per arwen and juliane, we can change to using `#titleize` now instead of capitalize, which should hopefully address this request (and others). If they later decide that words such as 'and', 'or', etc. shouldn't be capitalized, we could consider writing a custom helper function.

hweng:
rails commit has been removed.

PR #178, opened 2016-03-16
Added subject facets like scientific name, lithology, series etc. DHH-793

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-793

### DIFF
--- a/lib/dams/dams_object.rb
+++ b/lib/dams/dams_object.rb
@@ -579,7 +579,6 @@ module Dams
 	      	
 	      # facetable topics
 	      insertFacets solr_doc, "subject_topic", load_topics(component.topic)
-	      insertFacets solr_doc, "component_#{cid}_subject_topic", load_topics(component.topic)
 	      insertFacets solr_doc, "subject_topic", load_temporals(component.temporal)
 	      
 	      insertFacets solr_doc, "subject_topic", load_builtWorkPlaces(component.builtWorkPlace)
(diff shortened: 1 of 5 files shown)


spec/features/catalog_spec.rb
```diff
@@ -199,6 +199,83 @@
   end
 end
 
+feature "Search and browse facet links for scientific name" do
```
mcritchlow: this feature name is slightly misleading, as it's more focused on browsing facet values, and not about discovering them via search. It also covers more than scientific name, and appears to include all the custom faceted subjects.

I suppose one option is to include a test inside this feature for confirming the facets display on a search (in this case for "QE8iWjhafTRpc My Object" and then rename to something more generic. Alternatively, we could call this feature something like "Browse facet links for custom subjects" and then have a separate test for asserting the search discovery? Whatever you think makes the most sense.

lib/dams/dams_object.rb
```diff
@@ -597,6 +597,22 @@ def to_solr (solr_doc = {})
 	      insertFacets solr_doc, "subject_topic", load_series(component.series)
 	      insertFacets solr_doc, "subject_topic", load_cruises(component.cruise)
 	      	      	
+          # uncomment for subject anatomy
+          #insertFacets solr_doc, "subject_anatomy", load_anatomys(component.anatomy)
+          #insertFacets solr_doc, "component_#{cid}_subject_anatomy", load_anatomys(component.anatomy)
+          insertFacets solr_doc, "subject_common_name", load_commonNames(component.commonName)
+          insertFacets solr_doc, "component_#{cid}_subject_common_name", load_commonNames(component.commonName)
```
mcritchlow: Why do components need a separate insertFacets call? aren't component facet values loaded into the primary facet such as subject_common_name? I'm unclear why we need the component_cid_XXX versions.

lsitu: I am wondering why we need it either but just make it consistent with subject_topic https://github.com/ucsdlib/damspas/blob/develop/lib/dams/dams_object.rb#L582 in case we use it in the future. Do you want me to remove them all?

mcritchlow: well, i'm unclear what those method calls are actually doing. because they're running the insertFacets method and specifying a facet name that doesn't exist, or at least isn't created in the catalog controller. So, I'd like to think we have no use for them and could safely remove them. Perhaps you could verify by adding a test that has a component-level common name or scientific name, and then assert that a search that matches the record shows the facet on the results view? Or something like that.

lsitu: Yes, I think those component level facet field values can be retrieved by looping through components and we shouldn't utilized them at all. I'll remove them all and add tests for component level subject facet searching.

mcritchlow: :+1: sounds good

lsitu: @mcritchlow The nested component_attributes doesn't seem like to be working for component creation, with just an empty Component stub created. I have to load the example with common name from files to test for component level subject facet searching. Do you have any ideas regarding the nested attribute support for component_attributes?

mcritchlow: @lsitu i'm afraid not. I think the pattern we've always used is to load a fixture for complex object testing. so i guess that would be the easiest route in this instance? it might be a gap in support for nested attributes and the version of the hydra stack we're on, though i don't know that for sure.

lsitu: Okay, I'll separate the component level subject faceting test then.

lsitu: @mcritchlow I've removed those component level facet codes and added test for component level common name testing, along with those tests to verify objects from those subject facets links. And the feature test message is also corrected to be general.  See commit https://github.com/ucsdlib/damspas/pull/178/commits/2c75d50e28789ec12dad1bef185322eee606763b .

mcritchlow:
:+1:

PR #179, opened 2016-03-16
Feature/DHH-783

hweng:


app/views/shared/fields/_title.html.erb
```diff
@@ -34,11 +34,11 @@
         subtitleVal += (subtitleVal.blank? ? "" : ", ") + translationVariant
       end
       
-      if !result['value'].blank?
-         title = getFullTitle result
+      if !result['name'].blank?
```
mcritchlow: If I'm reading this right, we are now preferring the value in 'name' (authoritativeLabel) and using the getFullTitle method as a fallback option? This makes good sense to me, I'm just trying to confirm that this is what Arwen and Juliane agreed to. Arwen's most recent comment in the ticket leaves me slightly unsure. Also, since Bamboo is seemingly still not able to properly run CI, can you confirm the full suite passes for you?

hweng: @mcritchlow Yes the full suite passed for me when i run from my local. It seems like that Arwen agreed to use the value in 'name' (authoritativeLabel) as the first place.  Should I ping her again?

mcritchlow: @hweng no, that's ok. I just wanted to confirm that was your impression of the conversation. This looks good then! I'll get it merged.

PR #183, opened 2016-03-29
Integrated ShareNotify(DHH-788)

hweng:


hweng:
Some issue need to fix, will open another pull request.

PR #187, opened 2016-04-12
Added subfix to the facet titles for scientific name and common name. DHH-804

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-804


mcritchlow:
:+1:

PR #188, opened 2016-04-13
Added the newrelic gem for performance tracking. DHH-806

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-806

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,8 @@ gem 'nokogiri', '1.6.7.2'
 gem 'therubyracer', '~> 0.12.1', :platforms => :ruby
 gem 'uglifier', '~> 2.7.2'
 
+gem 'newrelic_rpm'
+
 group :development, :test do
   gem 'sass-rails', '~> 5.0.4'
   gem 'coffee-rails', '~> 4.1.1'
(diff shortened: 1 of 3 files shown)


mcritchlow:
gem inclusion looks great. 

it would be nice to get the config file sorted out as part of this.

here's a sample from one of Stanford's apps: https://github.com/sul-dlss/SearchWorks/blob/master/config/newrelic.yml

@rstanonik do you have the config from NewRelic? in our current deployment config for damspas, that will probably need to go in private config.

lsitu:
@mcritchlow I've added the content for the sample yml newrelic.yml.sample. rstanonik may want to add the real file to share/config folder for privacy and create a symlink for it. If we go this way, do we still need to add the real newrelic.yml to private_config?

mcritchlow:
@lsitu i'd prefer that we treat this in the same way we treat all private configuration for this app. That will make migration to the new model Vivian is working on much easier. So, let's please go that route. Once you have it setup in private config please let me know, and I'll merge this. 

Also, I'm not sure whether you want Ron's commit on the PR or not.

lsitu:
@mcritchlow I've added the yml file to private_config: https://lib-stash.ucsd.edu/projects/ND/repos/private_config/commits/c4643ab7562f8438ab1b9ca8b03c90e0d1f6d5cd

mcritchlow:
:+1:

PR #190, opened 2016-04-15
Enable the newrelic gem.

lsitu:
Clean up the gemfile lock.


mcritchlow:
we need for the NewRelic gem to be included in production for this release, and it is currently commented out in the Gemfile. It looks like Ron did that directly.

@lsitu - Can you please uncomment so we can get it built and tested out on staging and eventually prod?

lsitu:
Yes. I've made it but the PR header doesn't make sense.Does it matter?

mcritchlow:
You should be able to edit the title of the PR in the Github UI

lsitu:
@mcritchlow: Done!

mcritchlow:
:+1:

PR #192, opened 2016-04-28
Added component pager to COV (DHH-801)

dt-ucsd:


mcritchlow:
@dt-ucsd this looks great! :clap: 

it would be nice if we had some basic functional test coverage, other than demonstrating the presence of the pager.  Examples:

- If you are at the min-index and click the previous button, nothing should happen (or at least no error)
- If you are at the max-index and click the next button, nothing should happen (or at least no error)
- If a user clicks previous, and there is a component at index-1, that component is loaded
- If a user clicks next, and there is a component at index+1, that component is loaded

PR #198, opened 2016-05-12
Refactoring file use code.(DHH-811)

hweng:


app/helpers/dams_objects_helper.rb
```diff
@@ -540,32 +522,34 @@ def grabPDFFile(parameters={})
...
+    fieldData = file_data(parameters)
 		result = nil
 
 		if fieldData != nil
 			fieldData.each do |datum|
 				files = JSON.parse(datum)
 				if files["use"].end_with?("-service")
 					result = files["use"]
-					if(result == "document-service")
+					if(result == "document-service" ||result == "video-service" ||result == "audio-service")
```
mcritchlow: These conditionals are repeated almost exactly in the grabServiceFile method. Is there a refactoring opportunity here?

hweng:
@mcritchlow The refactored code use file format and use metadata no matter what the order of the file name, and it would render the correct format even there are two service file.  There is some difference between the grabFileUse() and grabServiceFile() methods, the former pull the data such as "video-service" and rendering the icon format and the latter pull the data such as "2.mp4" which would appended to the url.  I have extract the boolean condition code to the common method of required_file_use?()

hweng:
All common code has been extracted to file_use() method.

mcritchlow:
👍 great refactoring. thanks @hweng !

PR #200, opened 2016-05-14
Fixed the spellcheck suggestions format error from SOLR 5. DHH-814

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-814

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -276,7 +276,15 @@ class CatalogController < ApplicationController
       end
       (@response, @document_list) = get_search_results
 
-	  spelling_words = @response.spelling.words
+      spelling_words = []
+      @response.spelling.words.each do |word|
+        # handle Hash format for solr 5
+        if word.is_a?(Hash)
+          spelling_words << word["word"] if !word["word"].nil?
+        else
+          spelling_words << word
+        end
+      end
 	  if(@document_list.size == 0 && params['spellsuggestions'].nil?)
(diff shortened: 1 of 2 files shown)


spec/controllers/catalog_controller_spec.rb
```diff
@@ -23,6 +23,7 @@ def assigns_response
     end
   
     it "should have search results for a poor query with multiple key words" do
+      sign_in User.create({:provider => 'developer'})
```
mcritchlow: why was this `#sign_in` addition needed for this test?

app/controllers/catalog_controller.rb
```diff
@@ -276,7 +276,15 @@ def index
       end
       (@response, @document_list) = get_search_results
 
-	  spelling_words = @response.spelling.words
+      spelling_words = []
+      @response.spelling.words.each do |word|
+        # handle Hash format for solr 5
```
mcritchlow: will this work for Solr 4 and 5, or just 5?

spec/controllers/catalog_controller_spec.rb
lsitu: @mcritchlow  I think the test object is created in curator view and this test is examining the the search result, while the other test won't.

app/controllers/catalog_controller.rb
lsitu: It'll work for both SOLR 4 and SOLR 5 as the spellcheck words will be converted to a string array for solr 4 and solr 5.

PR #206, opened 2016-05-24
Feature/DHH-819

hweng:


spec/features/dams_collections_spec.rb
```diff
@@ -215,7 +215,7 @@
     scenario 'should see SHARE output' do
       sign_in_developer
       visit osf_api_dams_collection_path @provCollection.pid
-      expect(page).to have_content('"Test Title : "')
+      expect(page).to have_content('"Test Title"')
```
mcritchlow: could we possibly add a test that includes a translationVariant title example? that might help with coverage of the change you made, for future edits/change requests.

hweng:
Sure. I will add it. But it has to create a separate testing record and testing case.
This test is for the case that the record title only have the main title value and the translationVariant value is blank. The expected result is that the colon shouldn't get appended.

mcritchlow:
@hweng sounds like a plan. thank you!

hweng:
@mcritchlow More tests have been added for share mapping title "2f3c473", Since there are several other commits between this one and the one last pushed "b85a083". So I don't think we want to squash them into one.

mcritchlow:
@hweng agreed, this looks good and the tests passed 👍

PR #207, opened 2016-05-25
Added curator id to the file urls for dams statistics. DM-146

lsitu:
Added curator id to the file path and object url to track curator access for RDCP statistics, https://lib-jira.ucsd.edu:8443/browse/DM-146.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -187,5 +187,13 @@ module ApplicationHelper
       return JSON.parse(datum)['name']
     end
   end
-    
+
+  def to_stats_path (path)
+    if !path.blank? && !session[:user_id].blank?
+      path += path.index('?').nil? ? '?' : '&'
+      path + 'access=curator'
+    else
+      path
+    end
+  end
 end 
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -128,6 +128,8 @@ module CatalogHelper
       url = file_path parts[0], "_" + parts[1].gsub("/","_")
     end
 
+    url = to_stats_path url
+
     image_tag( url, :alt => "", :class => 'dams-search-thumbnail') unless url.blank?
   end
 
--- a/app/views/catalog/_document_thumbnail.html.erb
+++ b/app/views/catalog/_document_thumbnail.html.erb
@@ -4,6 +4,8 @@
     else
       url = dams_object_path(document, :counter => opts[:counter] )
     end
+
+    url = to_stats_path url
     
     restrictedNotice = grabRestrictedText(document['otherNote_json_tesim'])
 %>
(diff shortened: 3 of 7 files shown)


mcritchlow:
I have a couple questions on this.

- Why are we passing the user id to the repository? I looked back at the ticket, and I don't see anything in the request asking for specific curator ids. In fact, they want to exclude all curator stats, so it seems like we're passing along information that they want excluded from the statistics being gathered for RDCP (non-curator access). If I missed something in looking back at the ticket, please let me know, and apologies.
- Do we need to pass this (or any) information to the repository? Could we flag/log the curator vs non-curator requests from the hydra application directly?

lsitu:
@mcritchlow For question #1,  think the problem here is that we need a way to identify the curator requests of files download and object views, which will be logged in the apache web log located at /pub/data1/weblog. We need to count these curator access for dams statistics. While for RDCP stats, I'll exclude them since we only need non-curator access to be counted.
For the second question, I think we don't need to pass it to the repository but just need to be logged the URL in the apache web log. Flagging/logging the curator vs non-curator requests from the hydra application directly may not help.

mcritchlow:
1) agreed, but my question was in regards to why we needed to pass the specific curator making the request. If we're being asked to log specific user requests, then this works great. But, I didn't see any mention of that in the ticket. Could we log the role ('curator', 'public') instead, for example?

2) so is the concern that the hydra app will log to log/production.log and you would have to crawl from there and the apache weblog? that's definitely valid, and would be a bit annoying to manage. can we have the hydra app log to apache, or another log (syslog?) that's easier to query/parse? I'm mainly curious about the need to introduce the dependency on DAMS Repo for this. It seems like the Hydra app has all the information it needs to log the requests, short of where it's actually logging to?

lsitu:
For #1, yes, I think we can just log the role (curator, public) instead.
For #2, I think we can ask Ron for the logging. But it looks like the weblog just logs any client request and  so there should be no dependency on DAMS Repo. I believe it's just a marker in the URL. Yes, I think it'll work if we want to explore the logging from the Hydra app. My only concern is that how to remove the overlapping logging requests that logged by apache and the hydra app. Would you like me to add Ron in to discuss the location of writing log for the Hydra app?

mcritchlow:
Sure, if you wouldn't mind go ahead and get his thoughts. I think that would be great. Perhaps there's a better way we could handle Rails logging. If you both determine it's not worth the effort, for now, then we'll use the current solution. Thanks for your patience with my questions!

lsitu:
Hi Matt,
I haven't got any message from Ron yet. I double check the damspas code, and I see it won't pass the user id to the repository at all. So the userid in the URL is just a marker and there are no dependency to any other apps. I think we can just change it to use something like access=(curator | public) instead of the userid. If this is really a bothering issue, maybe we can try the ip way to excluded the staff workstation ip and vpn ip range instead?

mcritchlow:
By dependency, I just meant that both apps would have to be aware of the fact that a new parameter is being passed. But point definitely taken on that. 

👍  to switching to using a role

let's give Ron a little time to respond. if he doesn't by later today we'll move forward. thanks!

lsitu:
Matt, I've changed it to use a role, access=curator.

PR #208, opened 2016-05-27
Update config files for new deploy solution. DHH-807

VivianChu:
Add circleci config. DHH-807

Add devise config. DHH-807

Add secret token config

Update env variables to have default value. DHH-807

Update circle ci config to improve the performance. DHH-807

Setup parallelism to improve test suites performance. DHH-8-7

Fix syntax error

Add file config for circleci parallelism. DHH-807

Remove parallel feature from circleci.

Update config files with new env variables name. DHH-807

Remove secrets.yml. DHH-807

Change the deploy branch for testing.

Change the deploy branch for testing. DHH-807

Add support for different env variables in share_notify.yml. DHH-807

Fix the deploy execution expired error. DHH-807

Comment out sitemap and rspec task

Update rspec command

Update staging deploy config to remove sitemap role

Update circleci and deploy config files. DHH-807

Update config files for new deploy solution. DHH-807

Add secret token config

Update env variables to have default value. DHH-807

Update circle ci config to improve the performance. DHH-807

Remove parallel feature from circleci.

Update config files with new env variables name. DHH-807

Remove secrets.yml. DHH-807

Change the deploy branch for testing.

Add support for different env variables in share_notify.yml. DHH-807

Fix the deploy execution expired error. DHH-807

Comment out sitemap and rspec task

Update rspec command

Update staging deploy config to remove sitemap role

Update circleci and deploy config files. DHH-807

Update circleci to use phantomjs. DHH-807

Update circleci config. DHH-807


mcritchlow:
@ucsdlib/developers please review this significant PR from Vivian to move damspas to the new deploy solution. Please add any comments or concerns. If and when this looks ready to merge, please 👍

app/controllers/dams_resource_controller.rb
```diff
@@ -188,7 +190,22 @@ def solr_escape (str)
     pattern = /(\+|\-|\&\&|\|\||\!|\(\)|\{\}|\[|\]|\^|\"|\~|\*|\?|\:|\\)/
     str.gsub(pattern){|match|"\\"  + match} 
   end
-
+  
+  def replace_config_file (file_name)
+    copy_config_file (file_name)
+    config = ERB.new(IO.read("#{APP_ROOT}/config/#{file_name}")).result(binding)
```
hweng: @VivianChu @mcritchlow: Looking at this function, it makes the config file loaded each time click the the push button or any other request which use it, Should we cache it?

VivianChu:
@hweng I made it like that to make the change to the file on the fly, and change the config file back to use correct share_notify.yml since we have different values. The way share_notify.yml use env variables are different than other yml.  If you have a better way to cache it, maybe we can try to do that on next sprint. Thanks

hweng:
@tchu:  Ruby has cache operator using ||= , please change it to:
 config ||= ERB.new(IO.read("#{APP_ROOT}/config/#{file_name}")).result(binding)

when u load it, if this is loaded prior, statement will not try to evaluate right part of equation.

VivianChu:
@hweng I added the cache operator to it and submitted my change. However, I tested it locally and I don't think the cache helps much because that file will get overwrite again in this line.
https://github.com/ucsdlib/damspas/blob/feature/new_deploy/app/controllers/dams_resource_controller.rb#L234

VivianChu:
@hweng I think if I can make it not overwrite the file if it's in production env., then it will help more.  Let me try that and submit my change again.  Thanks

hweng:
@VivianChu  Yes,  I agree it is better to get rid of those redundant code in #copy_config_file, #replace_config_file and make it works in more logical way.

The cache operator is helpful as it makes your code to avoid re-evaluate the right part of equation, in this case it is the config file which makes much more sense to use this operator.

I used the following three liner approach when integrating the hydra-ldap to ILL-Billing several months ago and it works good with either private_config or Env:
https://github.com/ucsdlib/billing-ill/blob/develop/app/models/user.rb#L74

Hope it is helpful!

VivianChu:
@hweng  The way you used to loaded yml is similar to the way I found for active fedora when we discuss about this issue.  Since I couldn't update the way shared api load yml, I had to find a way to work around from our code.
https://ucsdlibrary.slack.com/archives/its-developers/p1461790708000133

hweng:
@VivianChu Ok, I will check the share_notify gem next sprint, and if it is the share_notify gem issue I would like to talk to Adam because the gem is supposed to work for the environment variable.

And in terms of our local code, my only concern for now is that:
1. It should use cache operator instead of regular operator for your case loading config file. I don't think it is good practice to re-evaluate the right side to load config for each http request. 
2. It's better to get rid of those redundant code in #copy_config_file, #replace_config_file.

VivianChu:
@hweng I agree with #1.  I know #2 is not a best way to handle it. If share_notify gem uses the similar way to load like Active fedora, then we can get rid of copy_config_file and replace_config_file. FYI - here's what I found days ago about their different way to load yml https://ucsdlibrary.slack.com/archives/its-developers/p1461792003000136

app/controllers/dams_resource_controller.rb
```diff
@@ -188,7 +190,22 @@ def solr_escape (str)
...
+    File.open("#{APP_ROOT}/config/#{file_name}", "w") do |f|
+      f.write(config)
+    end
+  end
+  
+  def copy_config_file (file_name)
+    data = File.read("#{APP_ROOT}/config/#{Rails.configuration.share_notify_sample}")
```
hweng: @VivianChu Can you add cache to this line, please change it to:
data ||= File.read("#{APP_ROOT}/config/#{Rails.configuration.share_notify_sample}"

app/controllers/dams_resource_controller.rb
```diff
@@ -188,7 +190,22 @@ def solr_escape (str)
...
+    File.open("#{APP_ROOT}/config/#{file_name}", "w") do |fi|
+      fi.write(data)
+    end
```
hweng: @VivianChu Is this repeated block of code as it is in #replace_config_file? I'm not sure why do you need it?

VivianChu: @hweng yes, I need it.  Here's the logic of this code, the first time it calls replace config file function. 

https://github.com/ucsdlib/damspas/blob/feature/new_deploy/app/controllers/dams_resource_controller.rb#L230
It will use the rails.config env to copy the yml file with correct env variables - either share_notify.yml.prod.sample or share_notify.yml.staging.sample  into the share_notify.yml.  The share_notify.yml will contain for example this "ENV.fetch('APPS_DHH_STAGING_SHARE_NOTIFY_HOST')". Then it will load the correct environment values, and it will write that data into itself share_notify.yml.  Now if you view share_notify.yml , it will contain the correct value for example https://staging.osf.io and token value instead.  After share_notify gem used it, if we don't copy back the sample one, then that become different file, git status will tell us that it is modified and if we commit that in, then that file no longer contains the env variable to be replaced and it contain our secret env variable values. For production, it's ok to not call this line, since we don't commit change from prod. However, for other environments we do need that.
https://github.com/ucsdlib/damspas/blob/feature/new_deploy/app/controllers/dams_resource_controller.rb#L233

hweng:
@VivianChu 
First I'm glad that you added cache, which improve the performance.

Second, what I meant the repeated code is that the following block of code repeated in #copy_config_file and #replace_config_file, is that better to move it to a method:
File.open("#{APP_ROOT}/config/#{file_name}", "w") do |f|
  f.write(config)
end

And We could figure out if there is a better approach for loading config file next sprint.

VivianChu:
@hweng Yes, I agree.  Thanks for all the feedback.

VivianChu:
@mcritchlow   Refactoring out that code currently gave me 14 errors.  I think we can either close this PR and carry it to next Sprint or open a new ticket for this?  What do you think? Thanks

lsitu:
👍

config/share_notify.yml.prod.sample
```diff
@@ -0,0 +1,3 @@
+---
+host: <%= ENV.fetch('APPS_DHH_PROD_SHARE_NOTIFY_HOST') {'https://staging.osf.io'}%>
```
mcritchlow: is this file needed?

config/share_notify.yml.staging.sample
```diff
@@ -0,0 +1,3 @@
+---
+host: <%= ENV.fetch('APPS_DHH_STAGING_SHARE_NOTIFY_HOST') {'https://staging.osf.io'}%>
```
mcritchlow: is this file needed?

config/share_notify.yml
```diff
@@ -0,0 +1,3 @@
+---
+host: <%= ENV.fetch('APPS_DHH_STAGING_SHARE_NOTIFY_HOST') {'https://staging.osf.io'}%>
```
mcritchlow: @VivianChu @hweng does this need to be broken up by environment since there are difference URLs/tokens in staging vs prod? Similar to config/solr.yml, for example.

config/share_notify.yml.prod.sample
VivianChu: @mcritchlow  Yes, that file is needed to copy to the share_notify.yml file depend on the environment like prod or staging.  It uses the correct environment variable name.

mcritchlow: Ok, I suppose this is related to my[ question about the share_notify.yml file](https://github.com/ucsdlib/damspas/pull/208#discussion-diff-65209334) I'm curious then, why we're not using the environment convention in the yml file for this, as we are with the other config files?

config/share_notify.yml
VivianChu: @mcritchlow  Yes, that's why I created the sample file share_notify.yml.prod.sample and share_notify.yml.staging.sample to make it use the correct env variables. 
Since the implementation of loading the env for share gem is different than active fedora gem, I can't just make it work an easy way like solr.yml or fedora.yml.  Here's the slack messages that we discussed before about this issue  https://ucsdlibrary.slack.com/archives/its-developers/p1461792003000136

config/share_notify.yml.prod.sample
VivianChu: @mcritchlow   If share api is updated to use the similar way like active fedora and Huawei's way to load the yml, then I can just make it use different env variables like other yml.  @hweng  Could you bring it up to Adam or whoever work on share api?  Thanks

config/share_notify.yml
mcritchlow: @VivianChu - ah, ok. so this is what's needed for the SHARE gem to work as-is? In that case, I'm 👍 for this now but noting that we should ideally refactor this, assuming @hweng can work with the other SHARE devs to make it possible. @hweng - should we setup a ticket in this Sprint for you to do that?

PR #215, opened 2016-06-08
Implemented new rules for complex object icons in search result page.…

hweng:
…(DHH-816)

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -144,7 +144,7 @@ module CatalogHelper
     resultClass = 'thumb-simple'
     resultIcon = 'glyphicon-file'
 
-    if document['files_tesim'] != nil
+    if document['component_count_isi'].blank? && !is_collection?(document)
       # SIMPLE OBJECT: look for preview image or format-appropriate icon
       field_data = document['files_tesim']
       simpleUse = file_use_type (field_data)
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -145,40 +145,43 @@ module CatalogHelper
     resultIcon = 'glyphicon-file'
 
     if document['files_tesim'] != nil
-
       # SIMPLE OBJECT: look for preview image or format-appropriate icon
-      fieldData = document['files_tesim']
-      simpleUse = nil
-
-      if fieldData != nil
-        fieldData.each do |datum|
-          files = JSON.parse(datum)
-          if files["use"].end_with?("-service")
-            simpleUse = files["use"]
-            break
-          end
-        end
-      end
+      field_data = document['files_tesim']
+      simpleUse = file_use_type (field_data)
       resultClass = 'thumb-simple'
       resultIcon = grabGlyphicon(simpleUse)
-
-    elsif document['component_count_isi'] != nil
-
-      # COMPLEX OBJECT: generic container icon
-      if document['component_count_isi'] > 0
+    elsif document['component_count_isi'] && document['component_count_isi'] > 0
+      # COMPLEX OBJECT: Folder icon or format-appropriate icon
+      types = document['resource_type_tesim']
+      if types && types.length >= 2
         resultClass = 'thumb-complex'
         resultIcon = 'glyphicon-folder-open'
+      else 
+        field_data = document['component_1_files_tesim']
+        simpleUse = file_use_type (field_data)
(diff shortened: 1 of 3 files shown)


hweng:
@mcritchlow I've extracted the existing block of code for file_use from catalog_helper.rb to make it shareable between the single object and complex object. The codeclimate caught there is a similar block of code in another class dams_object_helper.rb.  Maybe we could refactor those classes to move the common code to the application_helper.rb? But I'm not sure if it will break some other features. Since the refactoring is beyond the scope of this ticket. Could you create another ticket for that task? Thanks!

mcritchlow:
Yes, I've setup another ticket for additional refactoring. Thanks!

Otherwise, 👍

VivianChu:
👍

app/helpers/catalog_helper.rb
```diff
@@ -145,42 +145,45 @@ def document_icon( document )
...
-        fieldData.each do |datum|
-          files = JSON.parse(datum)
-          if files["use"].end_with?("-service")
-            simpleUse = files["use"]
-            break
-          end
-        end
-      end
+      field_data = document['files_tesim']
+      simpleUse = file_use_type (field_data)
```
lsitu: I am not sure whether it's sufficient to determine a simple object with the files_tesim field and a -service use of a existing file or not. There is a use case that DOMM user would like to add files to the top level of a complex object. A similar case they've done so far is just ingest a file in the first component, and along with it are other child components, see http://library.ucsd.edu/dc/object/bb1162536b (curator access) . Otherwise 👍 .

hweng:
@lsitu The ticket scope is "For complex object of one format type, use the format specific icon instead of the folder icon in search results".  
For simple object icon code that you mentioned above I didn't touch it because there is the risk to break the current feature.  If you think there is some new or special records that don't work with our code. Maybe we could request to update metadata to make them be more consistent? I think adding too many conditions would make the code unnecessary complicated.
And  I'm not sure how much inconsistent in our current code. So what i did is trying not to change bigger scope which might break something.  Thanks for comments!

lsitu:
Maybe it's safer to first handle the complex object case then the simple objects:
if document['component_count_isi'] && document['component_count_isi'] > 0
...
elsif document['files_tesim'] != nil
...
end

But I am fine to leave it for future refactoring. Thoughts?

hweng:
That would be fine for me.  Or we can use component_count_isi to detemine if it is complex object or simple object, is that safer?
if document['component_count_isi'] && document['component_count_isi'] > 0
...
elsif document['component_count_isi'].blank? && !is_collection?(document)
...
elsif is_collection?(document)

lsitu:
I think it'll work but I am fine to give it a 👍 and you can merged it, though I am feeling better if the order of the logic can be changed from special to common (simple objects as default) like:
if is_collection?(document)
elsif document['component_count_isi'] && document['component_count_isi'] > 0
else
#simple object
...

hweng:
Thanks for the comments. I think It is a little bit risk to have simple object as default.  It's safer to consider if there is some edge case which don't fall into these 3 category, then resultClass and resultIcon would fall into the default value which is assigned at beginning. The order doesn't matter to me as it clearly define the scope.
I pushed the new commits to reflect it.

lsitu:
@hweng Actually your new implementation may have included all those edge cases in you first logic:
if document['component_count_isi'].blank? && !is_collection?(document)
But well, I am fine. What do you think, @mcritchlow and @VivianChu?

VivianChu:
I'm fine with this implementation for now.  There could be room for improvement in the future.  @mcritchlow , @lsitu - is it ok to merge this PR now?

hweng:
@lsitu Logically there is the edge cases don't  included in the three categories, such as if document['component_count_isi'] = 1 or document['component_count_isi'] = 0 
I think these are not supposed to happen but who knows, so the safest way is to make those fall to default.

lsitu:
@hweng I think the case document['component_count_isi'] = 1 is real and we have some records like that and I think it will be caught by the complex object block. But I don't know how document['component_count_isi'] = 0 happened if it's not setting for all simple objects in general.

hweng:
@lsitu I'm not 100% sure what else code would definitely produce or what metadata would be either , so I think the safest is to make the code I'm digging on to work with every case as possible and not to throw the error. And I think for complex objects condition it makes more sense to be document['component_count_isi'] > 1 Thank you for all comments.

PR #216, opened 2016-06-09
Remove redundant subject types from Topic facet. DHH-809

VivianChu:


mcritchlow:
@ucsdlib/developers please review

lsitu:
👍 per the scope of the ticket https://lib-jira.ucsd.edu:8443/browse/DHH-809. Just note that there are 2 or 3 more subjects may be affected and need to removed from the topic facet: https://github.com/ucsdlib/damspas/blob/develop/app/controllers/catalog_controller.rb#L139-L145 .

VivianChu:
@lsitu Thanks for reviewing.  I noticed those facet fields before , but they didn't ask for them to be removed.

lsitu:
@VivianChu Right. Looks more like a future improvement.

mcritchlow:
👍

PR #221, opened 2016-06-14
separate dev and test data

mcritchlow:
@ucsdlib/developers - I think the reason why the development data that you're loading in dams-vagrant is getting wiped out after a run of the test suite is that the development and test database configurations were using the same database `db/development.sqlite3`

This change add an explicit test database. Can a few of you please:

- Pull down this branch into your dams-vagrant instance
- Load some sample records
- Run the test suite
- Check to see if your sample records are still available in the dev database

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -8,10 +8,10 @@ default: &default
   pool: 5
   timeout: 5000
   database: db/development.sqlite3
-  
+
 development:
   <<: *default
-  
+
 pontos:
   <<: *default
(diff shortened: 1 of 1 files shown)


VivianChu:
@mcritchlow I loaded an object locally and after running the test suites, that record got deleted.

mcritchlow:
@VivianChu - darn. ok, thanks for testing. perhaps there's another piece to this beyond having DB's separated from each other.

PR #226, opened 2016-06-30
Adding support for collection updates triggering share API updates.(D…

hweng:
…HH-843)

### DIFF
--- a/bin/damsolrizer
+++ b/bin/damsolrizer
@@ -101,12 +101,25 @@ begin
       max_tries = 2
       puts @msg.headers.inspect
       puts "PID: #{@msg.headers["pid"]}"
+      puts "method: #{method}"
       begin
-        if ["addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
+        if ["modifyCollection", "addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
           solrizer = Solrizer::Fedora::Solrizer.new
           solrizer.solrize @msg.headers["pid"], solropts
           puts  "updated solr index for #{@msg.headers["pid"]}\n"
-        elsif method == "purgeObject"
+          if method == "modifyCollection"
+            msg = Dams::ShareApi.share_push(pid)
+            if msg == "success"
+              puts  "Pushed #{pid} to OSF\n"
+            elsif msg == "failed"
+              puts  "The collection is not valid, #{pid} is failed to push to OSF\n"
+            elsif msg == "non-public"
+              puts  "The collection is not public, #{pid} is not pushed to OSF\n"
+            else
+              puts "System or service error, #{pid} is not pushed to OSF\n "
+            end
+          end
+        elsif method == "purgeObject" || method == "purgeCollection"
           ActiveFedora::SolrService.instance.conn.delete_by_id(pid)
           puts  "removed solr index for #{@msg.headers["pid"]}\n"
         elsif method == "testIndexer"
--- a/lib/dams/share_api.rb
+++ b/lib/dams/share_api.rb
@@ -1,0 +1,213 @@
+require 'open-uri'
+require 'httparty'
+
+module Dams
+  module ShareApi
+    
+    def self.share_push(pid)
+      env = Rails.env || 'production'
+      msg = ""
+   
+      begin
+        if env == 'production' ||  env == 'staging' || env == 'development'
+          hostname = Rails.configuration.host_name
+          doc = JSON.load(open("#{hostname}/collection/#{pid}/osf_data"))
+
+
+          if is_public?(doc) 
+          
+            document = ShareNotify::PushDocument.new("#{hostname}/collection/#{pid}", osf_date(doc))
+            document.title = osf_title(doc)
+            document.description = osf_description(doc)
+            document.publisher = osf_publisher
+            document.languages = osf_languages(doc)
+            document.tags = osf_mads_fields(doc)
+            osf_contributors(doc).each do |contributor|
+              document.add_contributor(contributor)
+            end
+            
+            if document.valid?
+              @headers = {'Authorization' => "Token #{share_config.fetch('token')}", 
+                         'Content-Type'  => 'application/json'
+                        }
+
+              @route = "#{share_config.fetch('host')}api/v1/share/data"
+              @response = with_timeout { HTTParty.post(@route, body: document.to_share.to_json, headers: @headers)}
+              msg = "success"
+            else
+              msg = "failed"
+            end
+          else
+            msg = "non-public"
+          end
+        end
+      rescue Exception => e
+        p e
+      end
+    end
+    
+    def self.is_public?(document)
+      !document.nil?
+    end
+ 
+    def self.share_config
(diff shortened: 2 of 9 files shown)


hweng:
@ucsdlib/developers - Please review, comment, etc... Thanks

bin/damsolrizer
```diff
@@ -101,12 +101,23 @@ begin
...
+      puts "method: #{method}"
       begin
-        if ["addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
+        if ["modifyCollection", "addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
           solrizer = Solrizer::Fedora::Solrizer.new
           solrizer.solrize @msg.headers["pid"], solropts
           puts  "updated solr index for #{@msg.headers["pid"]}\n"
-        elsif method == "purgeObject"
+          if method == "modifyCollection"
+            msg = Dams::ShareApi.share_push(pid)
```
lsitu: Will this push all collections with type "modifyCollection" to OSF? Do we need to check whether a collection is in OSF or not first before pushing the share api update?

mcritchlow: To add to @lsitu's question here, we should only be submitting "public" collections. We need to ensure UCSD-only and curator-only collections are not submitted.

lib/dams/share_api.rb
```diff
@@ -0,0 +1,203 @@
...
+      osf_data
+    end 
+
+    def self.osf_uris(document)
+      field_name = "id"
+      dams_data = document["#{field_name}"]
+      osf_data = {}
+
+      if dams_data != nil
+        url = "http://library.ucsd.edu/dc/collection/#{dams_data}"
```
VivianChu: @hweng - you already could access the field Rails.configuration.host_name , I'm wondering if it could be used here instead of the hardcode url. Same question for the line 186 
 +      osf_data = {"name": "UC San Diego Library, Digital Collections", "uri": "http://library.ucsd.edu/dc"}

hweng:
@mcritchlow Yes only "public" collections will be pushed to OSF. For non-public collections, it will give a message "Not pushed #{pid} to OSF".

mcritchlow:
@hweng - perfect, thanks for clarifying. What are your thoughts about including something in the message in the situation where a collection wasn't pushed to OSF because it is not a public collection?

hweng:
@mcritchlow Maybe a message like "Not pushed #{pid} to OSF due to non-public collection or some service error"? Let me know if you have some suggestion.

hweng:
@VivianChu The Rails.configuration.host_name get the current environment hostname, but in url field it is something like permanent link, must use http://library.ucsd.edu/dc/collection/".

VivianChu:
@hweng Thanks for clarifying.  👍

mcritchlow:
@hweng that sounds good. is it possible to distinguish between the two? So one message for when someone tries to submit a non-public collection, and then a separate error for the more general system/service error?

hweng:
@lsitu The order would be first the solr indexing got updated, then it call Dams::ShareApi.share_push.
The autoCommit does a hard commit to the disk. Let me know if there is any advantage you would think to use it in solr indexing?

hweng:
@mcritchlow Sure no problem. I will update the program to distinguish two.

hweng:
@mcritchlow I've updated program to distinguish between the non-public collection and general system/service error. Would your review again?

bin/damsolrizer
```diff
@@ -101,12 +101,25 @@ begin
...
+            if msg == "success"
+              puts  "Pushed #{pid} to OSF\n"
+            elsif msg == "failed"
+              puts  "Document is not valid, #{pid} is failed to push to OSF\n"
```
mcritchlow: I think this is great. At the moment since we're targeting collections, should we use `Collection` instead of `Document`? I'm just wondering if that will be more clear to end users.

hweng:
@mcritchlow It makes sense. Please feel free to edit the following messages.  All logical implementation has been added to share_api module so that the damsolrizer can keep clean. I've tested using damsolrizer-single on my local that it has been working properly.

if msg == "success"
     puts  "Pushed #{pid} to OSF\n"
 elsif msg == "failed"
      puts  "The collection is not valid, #{pid} is failed to push to OSF\n"
  elsif msg == "non-public"
       puts  "The collection is not public, #{pid} is not pushed to OSF\n"
  else
       puts "System or service error, #{pid} is not pushed to OSF\n "
  end

mcritchlow:
👍

bin/damsolrizer
```diff
@@ -101,12 +101,23 @@ begin
       max_tries = 2
       puts @msg.headers.inspect
       puts "PID: #{@msg.headers["pid"]}"
+      puts "method: #{method}"
       begin
-        if ["addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
+        if ["modifyCollection", "addDatastream", "addRelationship","ingest","modifyDatastreamByValue","modifyDatastreamByReference","modifyObject","purgeDatastream","purgeRelationship"].include? method
           solrizer = Solrizer::Fedora::Solrizer.new
           solrizer.solrize @msg.headers["pid"], solropts
```
lsitu: @hweng For your comment "@lsitu The order would be first the solr indexing got updated, then it call Dams::ShareApi.share_push.
The autoCommit does a hard commit to the disk. Let me know if there is any advantage you would think to use it in solr indexing?" --
I think currently setting of the Solrizer should be indexing with a soft commit. So I am wondering whether it's will do a hard commit for you or not. If it does a hard commit here, then the Solrizer is misconfigured, which slow down the indexing process a lot. So "objectModify" type and "collectionModify" type should be done in the other way with solropts={:autocommit => true | false}.

mcritchlow:
@ucsdlib/developers are there any remaining questions with this PR? If not, could someone please merge?

mcritchlow:
@lsitu - right. so we should test this out in QA properly then. we talked about this a bit as a group on Wednesday and @hweng is ready to do that. And, as you said, make sure that at least for this collection record context it uses hard commits?

hweng:
@mcritchlow In this pull request, nothing changed for solr indexing. We will test the soft commit then decide what will do the next step, right?

mcritchlow:
@hweng yes that sounds like a good plan. Based on the testing, we can determine next steps.

hweng:
@mcritchlow The SHARE push has been enabled in staging and production environments, but not qa. Maybe we can test it in staging?  From my local testing using damsolrizer-single, it worked as expected.
But not sure how it is going on in the staging and production environments which would depends on the damsmanager/damsrepo sending requests. And I've also put message for modifyCollection parameter to see if damsrepo send the correct parameter.

mcritchlow:
@hweng - sure thing. we can test on staging as part of the release review process. we'll just need to make sure we can queue up a number of items to index and test out the new functionality. please work with @lsitu and @VivianChu as needed for that.

lsitu:
@mcritchlow @hweng I think we need to deploy damsrepo to staging first.

mcritchlow:
@lsitu @hweng - we can do them both (all) as part of the release process I think. We need to do a release candidate for damsrepo anyway, right?

lsitu:
@mcritchlow Yes. But we can release it by its own with the index type changes.

PR #232, opened 2016-07-13
Fix random function issue. DHH-847

VivianChu:


lib/dams/random_controller_helper.rb
```diff
@@ -1,9 +1,9 @@
 module Dams
   module RandomControllerHelper
     # Retrieve random page from solr index
-    def random_page(q)
-      @response, @document = get_search_results(:q => "#{q}", :rows => 1)
-      page_count = @response.response['numFound']/20.to_f   
+    def random_page(q, fq)
+      @response, @document = get_search_results(:qf => "#{q}", :fq => "#{fq}", :rows => 100)
```
mcritchlow: overall 👍 

would it be faster to have a lower number of initial results returned, which I suppose would result in a larger number of pages to sample from? Do we need 100, or could we do something like 5 or 10?

VivianChu: @mcritchlow You're right, I forgot to change it back to 1 when doing testing.  
For that function, we just need to get the number of results found.  I will change it back to 1.

lsitu:
👍

VivianChu:
Thanks everyone for reviewing.  Could anybody please merge this PR?

PR #237, opened 2016-08-01
Improve log output with STDOUT logger and lograge with logstash format. DHH-844

lsitu:
https://lib-jira.ucsd.edu:8443/browse/DHH-844


lsitu:
@ucsdlib/developers Please review and comment. Thanks.

mcritchlow:
Do we have `LOG_LEVEL` defined anywhere in the config files? If not, should we? The fallbacks from the `fetch` call (`:info` for production, for example) look reasonable to me.

lsitu:
Yes, I don't think we've defined LOG_LEVEL in the config files but I assume the default value will taking effects. Would you like to add it to environment variables or just remove it and set it to the default value?

mcritchlow:
I think we should probably do one or the other. Either put it in the ENV, since the code in the PR is referencing it, or just update the code in the PR to use a static value. I'm open to whichever option makes sense, but as it the fetch call to ENV isn't really doing anything.

lsitu:
@mcritchlow I've move ahead to remove the lines for configuration with the environment variable LOG_LEVEL. So the default info log level should be applied for all environments now.

mcritchlow:
@VivianChu @hweng - any thoughts or comments?

VivianChu:
👍

VivianChu:
@lsitu I'm wondering how come we don't use debug log level for other environments except production.

lsitu:
@VivianChu Yes, I think by default only prod is info, while the default for other environment should be in debug mode, and this is what our system currently do. We can turn on it with environment variable setting later if it's not the case. Sorry about the confusing message.

VivianChu:
Thanks @lsitu for clarification.

hweng:
:+1:

lsitu:
Thank you all for reviewing and merging the PR.

PR #239, opened 2016-08-11
Fixing topic issue in index page.(DHH-854)

hweng:


hweng:
@ucsdlib/developers - please review, comment, etc... Thanks

VivianChu:
@hweng - Is it possible to have a test for this change?

hweng:
@VivianChu The test has been added.

app/controllers/catalog_controller.rb
```diff
@@ -191,7 +191,7 @@ def exclude_unwanted_records(solr_parameters,user_parameters)
     
     config.add_search_field ('Keyword (Title, Name/Creator, Topic, Notes etc.)') do |field|
 	  #field.solr_parameters = { :'qf' => 'all_fields_tesim' }
-	  field.solr_parameters = { :'qf' => 'title_tesim^99 name_tesim^20 subject_tesim^10 scopeContentNote_tesim^6 all_fields_tesim fulltext_tesim id' }
+	  field.solr_parameters = { :'qf' => 'title_tesim^99 name_tesim^20 topic_tesim^10 scopeContentNote_tesim^6 all_fields_tesim fulltext_tesim id' }
```
lsitu: Will it change the display order in the search result page for other subject types like genre, geographic etc?

hweng: @lsitu If use subject_tesim, the topics didn't show up in the search result page that's what i found the problem source described in the ticket:  See http://library.ucsd.edu/dc/search?f%5Bcollection_sim%5D%5B%5D=Dr.+Seuss+Political+Cartoons&id=bb65202085

Look at this line of code: config.add_search_field Keyword (Title, Name/Creator, Topic, Notes etc.),  Should we explicitly add this four solr_parameters only?

lsitu: @hweng I believe this is only a wording issue since we are doing fulltext searching.

lsitu:
Is the problem caused by search or using different fields in different places for UI display?

hweng:
@lsitu  Yes. Did you see the above record, the result page doesn't include the topic even the record does have topic if you use subject_tesim in config.add_index_field? This is the issue of this ticket and I have to change it to use topic_tesim to make the topic display on search result page to match to the object result page.

For config.add_search_field that you mentioned above, it is beyond this ticket scope so it is fine to me whether or not to update it. But i think it makes more sense to use topic_tesim instead of subject_tesim to match its word, and there is all_fields_tesim at the end anyway, right?

lsitu:
@hweng Maybe we just make the display fields for each component consistent for UI display since this is what the problem it is, but keep the search result as it is?

hweng:
@lsitu, I don't think this update would change the display order for other subject types like genre, geographic etc. But as it is beyond this ticket scope, I agree to keep that part of code as it was.

lsitu:
I recalled that all subject types should have the same priority for display ordering, this should include the following fields: https://github.com/ucsdlib/damspas/blob/develop/app/models/datastreams/dams_resource_datastream.rb#L804-L807 .

hweng:
@lsitu I reverted config.add_search_field to what it was before, as it is beyond this ticket scope.
The commit was squashed to f1c2a32.  
Regarding your last comment, It seems the topic doesn't show up if you use subject_tesim, and not sure if other subject type would show, there might be some issue with it which can be discussed in the future.

lsitu:
👍 It sounds good but it seems like there's the index issue there if it the topic doesn't show up.

VivianChu:
👍  Could you squash the commits and I will merge it for you? Currently I'm seeing 17 files changed.  Thanks

lsitu:
@hweng Does your commit carry my other commits and just trigger some changes to the github repository? It's strange that I just receive an automatic email from ticket https://lib-jira.ucsd.edu:8443/browse/DHH-844, which is opened automatically for code review: https://github.com/ucsdlib/damspas/pull/239#issuecomment-239250289 .

hweng:
@lsitu That's probably due to do rebase to combine the new commits to the older one. But no code was affected in any commit.

lsitu:
@hweng Interesting issue! Thanks for the clarification.

PR #266, opened 2016-12-13
Fixes #261 - Removed extra underscore from OG image URL

### LINKED ISSUE #261: Share feature and Facebook: complex object images do not get pulled in
When using the "AddThis" feature to share an image on a Facebook page, the following has been noted:

1. bb35511535: First page cover sheet is pulled in instead of the image.
bb35511535_page1

2. bb4199286c: Image for complex object is not pulled in (blank).
bb4199286c_image001

3. bb4029610p: Backside of postcard image (and front side) of the complex object is not pulled in.
bb4029610p_backside

dt-ucsd:
Removed superfluous underscore from the filename of the image in the
Open Graph image URL for complex objects.

### DIFF
--- a/app/helpers/dams_objects_helper.rb
+++ b/app/helpers/dams_objects_helper.rb
@@ -359,7 +359,7 @@ module DamsObjectsHelper
       while file == 'no_display' && i < component_count do
         i = i + 1
 	    file = grabDisplayFile(componentIndex: i)
-        file = "_#{i}_#{file}" if file != 'no_display'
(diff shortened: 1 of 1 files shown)


dt-ucsd:
I need help deciphering the circleci error. I don't have any experience with the tool so I don't know what's going on there. Thanks!

lsitu:
@dt-ucsd It looks like the accesslint key ACCESSLINT_GITHUB_USER is missing. @VivianChu, do you have any idea regarding the key?

PR #284, opened 2017-03-03
Clean up old form code

### LINKED ISSUE #282: Clean up old form code
### Descriptive summary

@VivianChu noticed that while we have replaced the contact form with an embedded MachForm, we haven't really cleaned up after the old form. 

### Expected behavior

Should remain as-is, but all code, DB migrations/table, gem for the old form should be removed.

### Related work

#275

VivianChu:
Fixes #282  ; refs #282

Clean up after the old form since we have replaced the contact form with an embedded MachForm.

* Remove contact_form model 
* Remove mail_form gem
* Add Test
* Remove some views that are not in in used


app/views/contact_forms/response.html.erb
```diff
@@ -1,4 +0,0 @@
-<h1>Contact Us</h1>
```
mcritchlow: I think this was needed for the MachForm, @dt-ucsd ?

config/routes.rb
```diff
@@ -5,10 +5,8 @@
   get '/p/:id', to: 'pages#view', :as => 'view_page'
 
   # Contact form routes
-  resources :contact_forms, :only => [:new, :create, :response]
-  post 'contact' => 'contact_forms#create'
+  resources :contact_forms, :only => [:new]
   get 'contact' => 'contact_forms#new'
-  get 'contact_response' => 'contact_forms#response'
```
mcritchlow: I think this was needed for the MachForm, @dt-ucsd ?

app/views/contact_forms/response.html.erb
VivianChu: @mcritchlow - Based on the Machform I created for DMR before, the response message was created with the Machform respond message option.

config/routes.rb
dt-ucsd: Yes. I was using that as the MF response while working on dev because the standard MF response wasn't working correctly. However on production the MF default reaponse seems to be okay. Wouldn't mind leaving it in as a backup if it is no bother.

app/views/contact_forms/response.html.erb
dt-ucsd: See reply below.

VivianChu:
@mcritchlow @dt-ucsd - The response form has been added.  Thanks

mcritchlow:
👍

lsitu:
👍

PR #292, opened 2017-03-28
Added the subject import and standard input steam Excel ingest templa…

remerjohnson:
The purpose of this PR is to add the "subject import" and "standard input stream" Excel templates. The reason we want to move them to `damspas` is because DOMM and RDCP should be able to share these templates with researchers or anyone inside (or outside) the library, which we cannot do currently because [dams-metadata](https://github.com/ucsdlib/dams-metadata) is private.

* Adding a directory in the root of the repo (`/ingest_templates/`), which hold the DAMS4 Excel templates


mcritchlow:
looks like CI timed out, rebuilding now. There might be an issue with pontos, if so we can ping OPS. Obviously this isn't going to break any tests :)

mcritchlow:
@lsitu or @hweng can you please check w/ @jhriv on pontos status?

lsitu:
@jhriv Ci/circleci build failed again. Have we made any change to pontos recently? When building for the last deployment, I already see it problematic (seemed to be volatile) though it succeeded finally. But it took more than two hours for a build, which is extreme slow. Could you take a look? I think we can restart everything to see how it goes if there are no change to it recently.

jhriv:
The only change to pontos recently that I am aware of is normal OS updates, and, the deleting of the failed deploy directory  for DMR. It is scheduled for a reboot to apply latest kernel updates. 

> Net::ReadTimeout

That is the only error I am seeing; following the backtrace I can't figure out what is going on there.

jhriv:
I triggered a rebuild on the last known good CircleCI build (the develop branch), and it, too, has failed with the `Net::ReadTimeout` error. Is CircleCI building in a Docker container, it looks like, so this is completely orthogonal to our pontos server.

My next attempt will be without the cache. If that fails, I'll spin up my own VM and run the tests.

mcritchlow:
would it be worth restarting the services on pontos that damspas needs to for running the test suite against?

jhriv:
I'm happy to reboot it, because there are kernel updates pending. If that's too disruptive, yes. we can restart the required services - if I knew what they were.

lsitu:
@jhriv Let's reboot it then since we don't know what's going wrong with the slowness and the volatile state that making it failed previously. Thanks.

jhriv:
Pontos has been rebooted; rebuilding the develop branch; initial signs look positive. When that completes, I'll kick off a build of this branch and we should be good to merge.

hweng:
@jhriv Just checked circleCI log, it looks fine to me so far comparing to the failed build.

jhriv:
... why was the build cancelled?

hweng:
I just cancelled another rebuild of develop branch because it was passed already. Then triggered the rebuild of add-ingest-template branch. As if running multiple build for damspas on circleCI same time, it would fail.

PR #330, opened 2017-07-06
Fix #325 - Adding Content-Type: image/jpeg for .jpg file to fix Calis…

### LINKED ISSUE #325: Run header check for images _2.jpg derivatives
### Descriptive summary

**Create a report:**
Run a script to crawl _2.jpg derivatives and look for those that have their **Content-type: application/octet-stream.**

Re-generate those derivatives to be **Content-Type: image/jpeg**

**Note:** Report should include the Collection Name so that Gabriela will know which collections to re-harvest at Calisphere.

### Rationale

Calisphere will not ingest/harvest our collections if it finds that an image has the wrong mime type (Content-Type: application/octet-stream)
```
HTTP/1.1 200 OK
Date: Thu, 15 Jun 2017 23:51:32 GMT
Server: Apache/2.2.15 (Red Hat)
Cache-Control: max-age=0, private, must-revalidate
Content-Disposition: inline; filename=bb0478952m_2.jpg
X-Powered-By: Phusion Passenger 5.0.30
Content-Length: 350401
Status: 200 OK
Content-Type: application/octet-stream
```

hweng:
…phere harvest issue.

### DIFF
--- a/app/controllers/file_controller.rb
+++ b/app/controllers/file_controller.rb
@@ -50,7 +50,9 @@ class FileController < ApplicationController
     disposition = params[:disposition] || 'inline'
     filename = params["filename"] || "#{objid}#{fileid}"
     headers['Content-Disposition'] = "#{disposition}; filename=#{filename}"
-    if ds.mimeType
+    if filename.include?('.jpg')
+      headers['Content-Type'] = 'image/jpeg'
+    elsif ds.mimeType
       headers['Content-Type'] = ds.mimeType
     elsif filename.include?('.xml')
       headers['Content-Type'] = 'application/xml'
--- a/app/controllers/file_controller.rb
+++ b/app/controllers/file_controller.rb
@@ -50,15 +50,15 @@ class FileController < ApplicationController
     disposition = params[:disposition] || 'inline'
     filename = params["filename"] || "#{objid}#{fileid}"
     headers['Content-Disposition'] = "#{disposition}; filename=#{filename}"
-    if filename.include?('.jpg')
-      headers['Content-Type'] = 'image/jpeg'
-    elsif ds.mimeType
-      headers['Content-Type'] = ds.mimeType
-    elsif filename.include?('.xml')
-      headers['Content-Type'] = 'application/xml'
-    else
-      headers['Content-Type'] = 'application/octet-stream'
-    end
+    headers['Content-Type'] = if filename.include?('.jpg')
+                                'image/jpeg'
+                              elsif ds.mimeType
+                                ds.mimeType
+                              elsif filename.include?('.xml')
+                                'application/xml'
+                              else
+                                'application/octet-stream'
+                              end
     headers['Last-Modified'] = ds.lastModifiedDate || Time.now.ctime.to_s
     if ds.size
       headers['Content-Length'] = ds.size.to_s


app/controllers/file_controller.rb
```diff
@@ -50,7 +50,9 @@ def show
     disposition = params[:disposition] || 'inline'
     filename = params["filename"] || "#{objid}#{fileid}"
     headers['Content-Disposition'] = "#{disposition}; filename=#{filename}"
-    if ds.mimeType
+    if filename.include?('.jpg')
```
mcritchlow: Assuming the team is OK with this strategy, this should come after the initial check for a valid mime type in the datastream.

As-is, it would preemptively assign the content type header for an image before even checking if we have a mime type from the technical metadata.

hweng: @mcritchlow The reason to place at the 1st place: if it is .jpg file, it would override the mime type in metadata, because for all .jpg file we want it set as Content-Type: image/jpeg regardless how it is defined in metadata, doesn't it?

lsitu: @hweng I agree with Matt. I think it may not be necessary to hardcode for the jpeg mime type here. The major issue here is that some of the .jpg derivatives generated in the past may be corrupted overtime for some reason, and we need to find out those corrupted .jpg derivative and fixed them, though sometimes it may be still displayable.

hweng: @lsitu It didn't say the jpg file is corrupted. 
It just say change Content-Type from application/octet-stream to image/jpeg.
The jpg file itself doesn't have MIME type. But it could set up in the damspas program, which it was set as application/octet-stream in else block.

lsitu: I think "application/octet-stream" is coming from the jpeg file itself unless the metadata record is corrupted., which was extracted by Jhove. If the format of a file can't be identified by Jhove, it will give it a mime type like "application/octet-stream" and a format like "bytestream".

hweng: @lsitu The example file in the ticket bb0478952m_2.jpg, It display fine that doesn't get corrupted.  I can open it without any problem. Do you get any problem for it?

The ticket require to change Content-Type from application/octet-stream to image/jpeg, which we need to fix here:
https://github.com/ucsdlib/damspas/blob/develop/app/controllers/file_controller.rb#L58

mcritchlow: The original concern I raised was that in preemptively checking the file type, we're bypassing our expected execution path of that method, which is that the value for the content type header is set by the mime type present in the technical metadata. This provides a promise, of sorts, that the kind of file we tell people we're sending is indeed what they're going to get. Only if that fails (there is no mimetype), do we fall back on the default `octet-stream`.

So I definitely agree with the statement "for all .jpg file we want it set as Content-Type: image/jpeg" but not with "regardless of how it is defined in the metadata". We don't want our metadata telling a different story than what we're telling downstream users. I see how that's kind of annoying in the case where an image renders just fine (the example in the referenced ticket), but it might not be the case in other instances, and we've had historical issues in that context, which I think is what @lsitu was getting at.

lsitu: @hweng I am assuming any jpeg derivatives we generated in the past that don't have a correct mimetype is corrupted in some extend, and we had better re-create them. Have you got a list of all jpeg derivatives that don't have a correct mimetype?

hweng: @mcritchlow The example file bb0478952m_2.jpg in https://github.com/ucsdlib/damspas/issues/325 doesn't get corrupted. It displays fine after I open it.   
The MIME is associated with record metadata, but I think the file itself doesn't contain it.

There are two options to get the Calisphere harvest working:
1. Fixing it in the program as what I did. I think Calisphere doesn't harvest metadata, does it?
2. Working with Ops to search the DAMS database and find all octet-stream for jpeg, and fix the metadata.

Which option do you prefer?


Do you prefer to update

mcritchlow: @hweng - Option 2 please. The current PR fixes the symptom (Calisphere getting files with the wrong content type header), but not the problem. The underlying problem is the mimetype is not available for those images, so that first `if` condition check that we expect to work, doesn't. Fixing the derivatives/technical metadata should in turn fix the symptom, the content type header being incorrect. We've done this kind of thing before, but in more targeted use cases. The hope here is to do something more broad.

Calisphere does harvest a lot of our metadata. I'm not sure whether they do anything explicitly with our technical metadata. But Calisphere aren't only people who are downloading our files, so we need a more global solution that addresses the underlying problem. I apologize if this could have been better clarified in the original ticket. Perhaps it was too focused on the specific example.

hweng:
@mcritchlow Sure, no problem.

app/controllers/file_controller.rb
hweng: @mcritchlow Sure, no problem. I will proceed it with option 2.
Do you get any case that the image got corrupted(can not be displayed) due to the wrong mimetype in its metadata?

mcritchlow: Thanks @hweng 👍

To answer your question: yep, that has happened in the past. @lsitu has dealt with that a few times. This is often with collections that were ingested many years ago, I believe.

_Hopefully_ re-running JHOVE will generate updated correct mimetype for most of the images, though re-creating some derivatives may be necessary.

lsitu: I agree. We need to re-create the derivatives for those that have a mime type like "application/octet-stream" for jpeg. This should be the right only way to fix the issue since we'll never know how other institutes like Calisphere utilize our jpeg images and the Jhove metadata on solr or extract the Jhove metadata from the image files themselves. Those files are damage in some extension though it could be open, otherwise we can't explain why they have the wrong mime type and format.

hweng: @mcritchlow Both Ron and I checked that the 2.jpg files itself don't contain header information such as MIME type, do they?
For http protocol, the header information is usually added in web application, that's why I used option 1 to override the MIME metadata from triplestore.
But since we decided to try the option 2:
If the derivative is fine that are not corrupted(like the example in ticket), re-creating derivative by image-magic would't fix the problem, but we should try to re-running JHOVE to fix the metadata in triplestore.  
Otherwise, both need to re-generate. 
But the side effect is that is there possible the derivative was fine but after re-generate it got corrupted? If it happened before, it could happen again.

lsitu:
@hweng We don't need to recreate the derivatives for all the image files but just those that are damaged. I think displayable shouldn't be the criterial to determine a corrupted jpeg derivative since we should have seen some cases that were truncated but still displayable in the past.

app/controllers/file_controller.rb
mcritchlow: Both Ron and I checked that the 2.jpg files itself don't contain header information such as MIME type, do they?

- They do not. We set them based on the extracted mimetype from JHOVE (and FITS for Horton).

If the derivative is fine that are not corrupted(like the example in ticket), re-creating derivative by image-magic would't fix the problem

- It might, actually. Because those derivatives were likely created on a much older version of imagemagick. 

but we should try to re-running JHOVE to fix the metadata in triplestore.
Otherwise, both need to re-generate.

- 👍 (for the same reason as above, many of them were likely run against an older version of JHOVE)

hweng:
@lsitu It's not only displayable, I also checked file size, image type and etc. properties use the get Info tool to determine if the file is corrupted.
What way do you use to define damaged file?

app/controllers/file_controller.rb
hweng: @mcritchlow I agree. I think is that possible some config issues in old JHOVE that get "octet-stream", assuming the derivatives created are fine.

lsitu:
If you look at all the derivatives in an objects like http://lib-hydratail-prod.ucsd.edu:8983/solr/blacklight/select?q=id:bb0478952m&fl=files_tesim, I don't know how to explain that only the 2.jpeg derivative have the wrong mime type "application/octet-stream" if it's not damage, while others like 3.jpg, 4.jpg, 5.jpg have the right format "JPEG" and mime type "image/jpeg".
@mcritchlow I recalled that the Jhove metadata were re-extracted during we migrated it to dams 4, so I think re-extract Jhove might not help in this case. But we can try.

mcritchlow:
That's some good background information, thanks @lsitu . Re-running JHOVE seems 'lighter' than new derivatives + JHOVE. But ultimately you might be right. I look forward to when we're using FITS :)

hweng:
@lsitu The link you posted is the metadata that we should fix. What I meant is that if the file is corrupted.
Can you just download the bb0478952m_2.jpg and open it and let us know if you think it is corrupted?

lsitu:
Yep, it's a historical issue. We'll get to it soon when moving forward to DAMS 5 ...

lsitu:
@hweng I don't know! I think this is a long long history and we'd regenerated the Jhove metadata during migrating to dams 4, which produced that "application/octet-stream" mime type. So both size and checksum won't work. We could just change the Jhove metadata if that is the way we want to do.

hweng:
Thanks for the background information, @lsitu 
Does the "application/octet-stream" issue only happen to _2.jpg? It seems the ticket only reported _2.jpg.

lsitu:
@hweng I don't know about the status of other JPEG derivatives. Maybe we should run a check for them as well? I think the major issue here is that why Callisphere report the JPEG files issue and how can we fix it. But even if the metadata part of a JPEG is damaged, it's still an issue.

hweng:
@lsitu My concern is that if it only happen to _2.jpg, it's more likely the old JHOVE config issue during migration to DAMS4? Then should try to fix it by only re-running the JHOVE.
But if it is randomly, it's more likely caused by file corruption, then it should re-running both .

lsitu:
I've never seen Jhove config is an issue for the JPEGs derivative that we generated and we seldom change it. Do you know how it's be related?

hweng:
I'm not sure how to explain "we'd regenerated the Jhove metadata during migrating to dams 4, which produced that "application/octet-stream" mime type" and the pattern if it only happen to _2.jpg, assuming the derivative files are fine.  JHOVE itself should be reliable, right?

lsitu:
Do you see it has any issues with the JPEG derivatives that we created? I won't argue about it but I think we've got a good chance to test it out while fixing the issue with ticket https://github.com/ucsdlib/damspas/issues/325.

PR #365, opened 2017-12-08
Fix the collection item view

### LINKED ISSUE #361: Distorted item view display for RDCP collection
### Descriptive summary

If using this URL to get to the item view display for the Southern Ocean Carbon and Climate Observations and Modeling collection, the item display is skewed starting from the third object down:

https://library.ucsd.edu/dc/search?utf8=%E2%9C%93&q=bb4473712z

### Steps to reproduce the behavior

- Go to DC Home Page and enter the ARK (bb4473712z)
- Note that you are taken to a results page that lists the collection record + items
- This odd display does not happen when you are in the collection record and you click on the View Items

### To Do

@VivianChu will ingest this collection onto the staging environment to better troubleshoot.

VivianChu:
Fixes #361  ; refs #361 

Fix distorted item view display issue for RDCP collection

* Remove f.note_tesim.hl.fragsize for note field
* Add fixture files and test

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,7 +19,7 @@ module ApplicationHelper
 		#Snippets for no default indexed fields
 		if(hitsonly && highlight_values != nil && highlight_values.count > 0)
 			highlight_values.collect! {|m|m.length < blacklight_config.hlMaxFragsize || m.ends_with?(".") ? m : m+ " ..."}
-			return highlight_values.join(sep).html_safe
+			return strip_tags(highlight_values.join(sep))
 		end
 		highlight_values = document[field] if (highlight_values.nil? || highlight_values.count==0)
 	elsif field.to_s.index('_json_')
--- a/spec/features/dams_collections_spec.rb
+++ b/spec/features/dams_collections_spec.rb
@@ -303,3 +303,44 @@ feature "Vistor wants to view the OSF API output" do
       expect(page).to have_content('{"name":"UC San Diego Library"}')
     end
 end
+
+feature 'Visitor wants to look at the collection item view search results' do
+  before do
+    @unit = DamsUnit.create(pid: 'bb45454545')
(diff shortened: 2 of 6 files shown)


VivianChu:
@mcritchlow - I tried to change the MaxFragsize to 250 before and it didn't fix the problem.  However, when I removed it, then it worked.

VivianChu:
@mcritchlow - attached is the screen shot when that value is set to 250.  However, when set it to 300, it also worked.  What do you think? Setting it to 300 or remove it?
note-search2

mcritchlow:
So after further discussion on the Sprint call, it seems like the issue is that we have embedded html in the Notes that are getting cut off in the highlight snippet. The challenge here is we'll never know exactly where these embedded tags might be in the note string, so I'm a little wary of setting a new fragsize value since it might break things for a different note.

Is is practical to strip out html from highlighted search result fields?

VivianChu:
@mcritchlow - https://github.com/ucsdlib/damspas/blob/develop/app/controllers/catalog_controller.rb#L84  , we can remove that and the tag will use the default value <em> instead and still cause the distorted view.  
Or we can remove note field from highlight feature - https://github.com/ucsdlib/damspas/blob/develop/app/controllers/catalog_controller.rb#L95

mcritchlow:
@VivianChu - sorry, i mean to say, can we remove the embedded html tags from the note field, not the span wrapper. Looking at the example you mentioned that's linked in the ticket, it looks like it's getting cut off right in the middle of rendering the link:

```
<span>For the latest snapshot of the float data, see the Related Resource section on the <a href="http://library.ucsd.edu/dc/collection/<span ...
```

If that's not practical (or we can't think of some other way to better control the html output of the highlighted snippet), then like you said we might have to consider something more significant, like not displaying those at all because there is so much variability in them..

VivianChu:
@mcritchlow - I'll try other way to remove the embedded html tags from the note field. I'll keep you posted.

PR #366, opened 2017-12-11
Fix distored collection items search view

### LINKED ISSUE #361: Distorted item view display for RDCP collection

VivianChu:
Fixes #361  ; refs #361 

Fix distorted item view display issue for RDCP collection

* Remove the embedded html tags from the note field
* Add fixture files and test

### DIFF
--- a/spec/features/dams_collections_spec.rb
+++ b/spec/features/dams_collections_spec.rb
@@ -334,13 +334,10 @@ feature 'Visitor wants to look at the collection item view search results' do
     @soccomObj2.delete
     @soccomObj3.delete
   end 
-  scenario 'should see the collection item result page with correct indentation' do
+  scenario 'should not see html tags for any hightlight field in the search result page' do
     visit catalog_index_path( {:q => "#{@aCollection.pid}", 'sort' => 'title_ssi asc'} )
     expect(page).to have_content('Showing results for 1 - 4 of 4')   
-    expect(page).to have_selector('div h3 a[data-counter="1"]', :text => 'SOCCOM float data - Snapshot 2016-12-28')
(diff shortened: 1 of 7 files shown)


VivianChu:
@ucsdlib/developers - please review.  
Here's the screen shot of the search result locally after removing the embedded html tags.
note_strip_tag

spec/features/dams_collections_spec.rb
```diff
@@ -302,4 +302,45 @@
...
+    solr_index (@soccomObj3.pid)       
+  end
+  after do
+    @aCollection.delete
+    @unit.delete
+    @soccomObj1.delete
+    @soccomObj2.delete
+    @soccomObj3.delete
+  end 
+  scenario 'should see the collection item result page with correct indentation' do
```
mcritchlow: Given the change to strip the html tags, I'm wondering if that's what this test should be more focused on. It's less about correct indentation (the side-effect) and more about assuring that any fields that we'd display as highlights in the search results don't contain html tags, is that right? I'm not sure if there's a more efficient way to test that specifically, off the top of my head, but maybe we should rename the scenario line?

PR #398, opened 2018-02-02
Downgrade Ruby version to 2.3 series

mcritchlow:
At the moment, we have our version of Rails pinned to the 4.2.7.x
version, which doesn't support Ruby 2.4. To fix this in the least
impactful way, we'll downgrade the version of Ruby to the latest in the
2.3 series. Going forward we could consider updating the Rails stack to [4.2.8+ which is the first to support Rails 2.4](http://weblog.rubyonrails.org/2017/2/21/Rails-4-2-8-has-been-released/)


hweng:
:+1:

mcritchlow:
Hey all, sorry for the poke but I'm hoping to get the dams-vagrant work wrapped up today as I'm starting a new project on Monday. @VivianChu or @lsitu would you mind merging when you have a chance? Thanks :)

PR #425, opened 2018-03-23
Fixes #410 - Add support for metadata-only to collection level view

### LINKED ISSUE #410: Collection level record view
### Descriptive summary

- As a user, in the metadata display section, I should see the Access states at the top.
- As a user, I should see the collection's image on the right and be able to View Collection Items.
- As a user, I should be able to Search this collection.

### Expected behavior

The access note for metadata-only collections should say:

**Access**
- **Restricted View**

### Example View

ece

### Reference

VivianChu:
Fixes #410  ; refs #410 

### DIFF
--- a/app/controllers/dams_resource_controller.rb
+++ b/app/controllers/dams_resource_controller.rb
@@ -40,6 +40,10 @@ class DamsResourceController < ApplicationController
     # generate facet collection list for collection page only
     models = @document["active_fedora_model_ssi"]
     if models.include?("DamsAssembledCollection") || models.include?("DamsProvenanceCollection") || models.include?("DamsProvenanceCollectionPart") 
+      other_rights_query = "(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay) AND collections_tesim:#{params[:id]}"
+      @res, @doc = get_search_results(q: other_rights_query, rows: 1)
+      @metadata_only = !@res.response['numFound'].zero?
+
         facet_collection_params = { :f=>{"collection_sim"=>"#{@document['title_tesim'].first.to_s}"}, :id=>params[:id], :rows => 0 }
         apply_gated_discovery( facet_collection_params, nil )
         @facet_collection_resp = get_search_results( facet_collection_params )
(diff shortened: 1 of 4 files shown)


app/views/shared/fields/_access_control_level.html.erb
```diff
@@ -7,11 +7,13 @@
 	if accessGroup != nil
 
 		if accessGroup.include?('public')
-		  viewAccess = nil
-	  elsif accessGroup.include?('local')
+		    viewAccess = nil
```
mcritchlow: what if instead of `nil` we called this `public`, or even make it a symbol? It seems like that's what nil is representing here.

Then below for conditonals you could say `if viewAccess != :public`, for example, instead of a nil check. There may be other downstream issues with this, just a thought.

app/controllers/dams_resource_controller.rb
```diff
@@ -40,6 +40,10 @@ def show
     # generate facet collection list for collection page only
     models = @document["active_fedora_model_ssi"]
     if models.include?("DamsAssembledCollection") || models.include?("DamsProvenanceCollection") || models.include?("DamsProvenanceCollectionPart") 
+      other_rights_query = "(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay) AND collections_tesim:#{params[:id]}"
+      @res, @doc = get_search_results(q: other_rights_query, rows: 1)
+      @metadata_only = @res.response['numFound'].zero? ? false : true
```
mcritchlow: Doesn't `.zero?` return true or false already? Do we need the ternary part? (` ? false : true `)

mcritchlow: Oh i see what you're doing here, you actually want the inverse. nevermind :) I suppose an alternative to the ternary is `@metadata_only = !@res.response['numFound'].zero?` but that's pretty subjective as far as what's more readable..

VivianChu: @mcritchlow I think your alternative is better.  I'll update it to use it.

app/views/shared/fields/_access_control_level.html.erb
lsitu: @VivianChu I have no idea how nil is used here but I believe the purpose is just not to show the label for the public view. So I think either way will be fine.

VivianChu: @lsitu - I used the symbol way that Matt suggested and the tests still passed locally.

PR #432, opened 2018-04-11
Add Access note to browse by collection page for metadata-only collection 

### LINKED ISSUE #409: Browse by collection view
### Descriptive summary

- As a user, I should see the collection's thumbnail and brief description.
- As a user, below the description, I should see all Access notes with the exception of "Curator Only"

### Expected Behavior

The access note for **metadata-only view** collections should say:

**Access**
- **Restricted View**

### Example View

collection

### Reference

VivianChu:
Fixes #409  ; refs #409 

Display the access note for metadata-only view collections.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,6 +28,10 @@ Style/Documentation:
     - 'spec/**/*'
     - 'app/**/*'
 
+Style/ConditionalAssignment:
+  Exclude:
+    - 'app/helpers/*'
+    
 Style/Next:
   Enabled: true
 
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -91,23 +91,21 @@ module CatalogHelper
     dateVal
   end
 
-  def display_access_control_level(document)
+  def display_access_control_level(document, metadata_colls)
+    access_group = document['read_access_group_ssim'] # "public" > "local" > "dams-curator" == "dams-rci" == default
+    view_access = 'Curator Only'
+    if access_group
 
-    accessGroup = document['read_access_group_ssim'] # "public" > "local" > "dams-curator" == "dams-rci" == default
-    viewAccess = nil
-
-    if accessGroup != nil
-
-      if accessGroup.include?('public')
-        viewAccess = nil
-      elsif accessGroup.include?('local')
-        viewAccess = 'Restricted to UC San Diego use only'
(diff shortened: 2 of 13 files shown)


VivianChu:

app/helpers/catalog_helper.rb
```diff
@@ -91,23 +91,23 @@ def date_list( document )
...
-
-    if accessGroup != nil
-
-      if accessGroup.include?('public')
-        viewAccess = nil
-      elsif accessGroup.include?('local')
-        viewAccess = 'Restricted to UC San Diego use only'
+  def display_access_control_level(document, metadata_colls)
+    access_group = document['read_access_group_ssim'] # "public" > "local" > "dams-curator" == "dams-rci" == default
+    view_access = nil
```
mcritchlow: It looks like the fallback value here, from the `else` condition, is `Curator Only`. If that's true, could you possibly set the initial value of `view_access` to that instead of `nil`? Then you could drop the `else` entirely.

Somewhat related I'm curious why we use `nil` instead of just `''` for `public`, but if it's working then I suppose it's not work messing with ;)

app/helpers/catalog_helper.rb
```diff
@@ -91,23 +91,23 @@ def date_list( document )
...
+    if !access_group.nil?
```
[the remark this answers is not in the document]
mcritchlow: +1, I think this could be `if access_group`, or simply an early return if it is nil.

.rubocop.yml
```diff
@@ -28,9 +28,17 @@ Style/Documentation:
...
+Style/ConditionalAssignment:
+  Exclude:
+    - 'app/helpers/*'
+    
 Style/Next:
   Enabled: true
 
+Style/NegatedIf:
+  Exclude:
+    - 'app/helpers/*'
```
mcritchlow: If the negated condition noted by Hound is dropped (turned into an else or early return/guard clause) will this still be needed?

VivianChu:
@mcritchlow - I updated the code that removed the negated condition from rubocop like you suggested and Hound didn't complain.  The circleci build is still running.  Thanks

VivianChu:
@mcritchlow - The code has been updated and the tests passed.  Could you review them again?  Thanks

mcritchlow:
LGTM :+1:

PR #442, opened 2018-05-01
Add generic thumbnail and restricted view info to search objects view

### LINKED ISSUE #411: Items view
### Descriptive summary

- As a user, I should see grey generic thumbnails for objects of metadata data-only visibility collections.

### Example View

objects

### Reference

VivianChu:
Fixes #411  ; refs #411 

Add grey generic thumbnails and restricted view access info for objects of metadata data-only visibility collections.

### DIFF
--- a/app/views/catalog/_document_thumbnail.html.erb
+++ b/app/views/catalog/_document_thumbnail.html.erb
@@ -8,12 +8,18 @@
     url = to_stats_path url
     
     restrictedNotice = grabRestrictedText(document['otherNote_json_tesim'])
+    access_notice = grab_access_text(document)
 %>
+
 <%= link_to url, :class => 'dams-search-thumbnail-link' do %>
   <%- if !restrictedNotice.nil? && rtn = restricted_object_url%>
     <div class="document-thumbnail dams-search-thumbnail">
       <%= rtn %>
     </div> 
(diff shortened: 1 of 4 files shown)


app/views/catalog/_document_thumbnail.html.erb
```diff
@@ -8,12 +8,18 @@
...
     restrictedNotice = grabRestrictedText(document['otherNote_json_tesim'])
+    access_notice = grab_access_text(document)
 %>
+
 <%= link_to url, :class => 'dams-search-thumbnail-link' do %>
   <%- if !restrictedNotice.nil? && rtn = restricted_object_url%>
     <div class="document-thumbnail dams-search-thumbnail">
       <%= rtn %>
     </div> 
+  <%- elsif access_notice && rtn = restricted_object_url%>
```
mcritchlow: How hard would it be to have a helper function that returns the URL we want for the link? We've got 3 branches here that all render the same div with the same classes and then the link. I'm just wondering if we could pull the logic out so the `link_to` is basically just the div with the returned value from a method that sorts out the correct URL. If that's a pain, then no worries! Just a thought.

VivianChu:
@mcritchlow - I created a helper method for the url.  Could you review them again?  Thanks

hweng:
:+1:

PR #453, opened 2018-05-17
Fixes #446 - Add integration test to replicate the object counts issue and fix the problem.

### LINKED ISSUE #446: Total number of objects no longer updates on DC home page.
### Descriptive summary

The total number of objects listed on the home page is no longer being dynamically updated. Also, the total is the same for public as well as when in curator mode (should show more in curator).

Currently totaling **117,586** research data sets, images, documents, video and audio recordings.

### Work needed

Look into why the code for gated discovery is no longer working.

lsitu:
Fixes #446

Fix the issue with the objects count that is the same for public user and curator with gated discovery not working.


VivianChu:
👍

mcritchlow:
Sorry this was never merged @lsitu :(

PR #454, opened 2018-05-17
Fixes - #270 Implemented mapping from DAMS4 to SHARE V2 API and updated push function.

### LINKED ISSUE #270: Implement new mapping and API for SHARE2 in DAMS4
This ticket will be visited after the following two tickets is finished:
share_notify gem support for SHARE2: https://github.com/ucsdlib/damspas/issues/268,
and share mapping spreadsheet rewriting: https://github.com/ucsdlib/damspas/issues/269

hweng:
Fixes #270 ; refs #270

### DIFF
--- a/lib/dams/controller_helper.rb
+++ b/lib/dams/controller_helper.rb
@@ -2,539 +2,596 @@ module Dams
   module ControllerHelper
     
     #Mapping for OSF API
-		def osf_title(document)
-			  field_name = "title_json_tesim"
-		    dams_data= document["#{field_name}"]
-		    osf_data=''
+    def osf_title(document)
+      field_name = "title_json_tesim"
+      dams_data= document["#{field_name}"]
+      osf_data=''
 
-		  if dams_data != nil
-		    dams_data.each do |datum|
-		      title = JSON.parse(datum)
-		      osf_data = title['name'] ? title['name'] : ''
-		      osf_data += title['name'] && !title['translationVariant'].blank? ? ' : ' : ''
-		      title_trans = title['translationVariant'] || []
-		      if title_trans.class == Array
-		      	title_trans.each do |trans|  
-		      		osf_data +=	trans
-		      	end
-		      elsif title_trans.class == String
-		      	osf_data +=	title_trans
-		      end
-		    end
-		  end
-		  osf_data
-		end
+      if !dams_data.nil?
+        dams_data.each do |datum|
+          title = JSON.parse(datum)
+          osf_data = title['name'] ? title['name'] : ''
+          osf_data += title['name'] && !title['translationVariant'].blank? ? ' : ' : ''
+          title_trans = title['translationVariant'] || []
+          if title_trans.class == Array
+            title_trans.each do |trans|  
+                osf_data += trans
+            end
+          elsif title_trans.class == String
+            osf_data += title_trans
+          end
+        end
+      end
+      osf_data
+    end
 
-		def osf_contributors(document)
-			field_name = "relationship_json_tesim"
-			dams_data = document["#{field_name}"]
-			osf_data =[]
+    def osf_contributors(document)
+      field_name = "relationship_json_tesim"
+      dams_data = document["#{field_name}"]
+      osf_data =[]
 
-			if dams_data != nil
-		    dams_data.each do |datum|
-		    	
-		      relationships = JSON.parse(datum)
-		    	relationships.each do |key, value|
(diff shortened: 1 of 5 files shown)


lib/dams/controller_helper.rb
```diff
@@ -44,6 +44,71 @@ def osf_contributors(document)
 		  osf_data = (osf_data.blank?) ? osf_data << {"name": "UC San Diego Library"} : osf_data
 		end
 
+		def osf_related_agents(document)
```
mcritchlow: Could you check the spacing on this file? It seems off/not aligned.

lib/dams/controller_helper.rb
```diff
@@ -44,6 +44,71 @@ def osf_contributors(document)
...
+			field_name = "relationship_json_tesim"
+			dams_data = document["#{field_name}"]
+			osf_data =[]
+
+			if dams_data != nil
+		    dams_data.each do |datum|
+		    	
+		      relationships = JSON.parse(datum)
+		    	relationships.each do |key, value|
```
mcritchlow: You could potentially use `map` here instead of generate the array for you. But just a thought, this definitely works as-is.

lib/dams/controller_helper.rb
```diff
@@ -44,6 +44,71 @@ def osf_contributors(document)
...
+		   		end
+		    end
+		  end
+		  osf_data << {agent_type: "Publisher", type: "Organization", "name": "UC San Diego Library Digital Collections"}
+		end
+
+		def agent_type(type)
+			type = 'principalinvestigator' if type == 'principal investigator' || type == 'Principal Investigator'
+
+			share_agent_type = [
```
mcritchlow: This looks like it could possibly go in a config file? Do you think there would be value in that?

lib/dams/controller_helper.rb
```diff
@@ -44,6 +44,71 @@ def osf_contributors(document)
...
+       ]
+      type = (share_agent_type.include? type) ? type : 'Contributor'
+		end
+
+		def osf_extra(document)
+      field_name = 'otherNote_json_tesim'
+      dams_data = document["#{field_name}"]
+      osf_data = {}
+
+			if dams_data != nil
```
mcritchlow: Could there perhaps be a guard clause here? Something like `return osf_data unless dams_data`?

lib/dams/controller_helper.rb
hweng: @mcritchlow The spacing for new added code looks fine to me from my local sublime editor which are all two space indent. Some old code such as #get_linked_object_label() and etc. are not aligned in this file. Do you want me to clean up those old code?

screen shot 2018-06-04 at 12 12 03 pm

screen shot 2018-06-04 at 12 13 15 pm

mcritchlow: Ah you're right. Those were getting mixed up in the diff. But your OSF-related methods do indeed look fine after I looked at them in a non-diff context. Sorry about that :)

hweng: Or maybe it is better to create a separate ticket to clean up and format old code if it is necessary?

mcritchlow: Personally I don't think it's the best use of your time. I think if/when we have to touch that code, we could use the opportunity to clean it up a bit. Just my opinion though :)

hweng: @mcritchlow  I found the reason why in the editor my code looks beautiful, but it looks far off in Github. That's because Github displays tabs as equivalent to 8 spaces. Now I update my editor setting to make the code in Sublime and Github look the same.  I can see the off space and will fix them from my editor. Sorry about that.

mcritchlow: Oh no worries. That actually was my original concern, that IDE "hide" spacing issues so one would never actually know that they exist. My editor (Vim) definitely does not do that ;)

Though this does make me wonder if we should write down what our expectations are for tabs/spaces and the spacing for them. Rubocop kind of enforces it, but we've also never written it down and we more or less are barely using Rubocop w/ damspas because it's got so much historical baggage. One thing I came across that I was going to run by you all is https://editorconfig.org/ which works with pretty much every text editor/IDE. Maybe I should ask in our slack channel?

hweng: Yes, sounds a good topic to discuss.

hweng:
Moved to https://github.com/ucsdlib/damspas/pull/460

PR #458, opened 2018-06-04
Only enable New Relic agent in Production

mcritchlow:
* Default new relic agent to be disabled
* Enable new relic agent on production *only*

The reason for this is we're going to start paying for New Relic, and we are charged per host. Which means having Staging, QA, etc. reporting data to New Relic is going to cost us an insanely large amount of money.. So we're going to focus on production. According to their documentation, this should do the trick.

@VivianChu - If this works we'll need to do the same for DMR

### DIFF
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -19,7 +19,7 @@ common: &default_settings
   app_name: damspas
 
   # To disable the agent regardless of other settings, uncomment the following:
-  # agent_enabled: false
+  agent_enabled: false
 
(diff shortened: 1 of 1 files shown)


mcritchlow:
@ucsdlib/developers - would you all be OK with merging this and deploying it at least on Pontos? That would let us remove at least one host from our New Relic costs. DMR is already off Pontos (thanks @VivianChu )

lsitu:
👍

VivianChu:
👍

Issue #467, opened 2018-06-20
Fix access controls for downloading mp3 and mp4 files


gamontoya:
### Descriptive summary

Users are able to download mp3 or mp4 files if they know the full URL path.

- /_2.mp3
- /_2.mp3?download

Objects (any type of content file) that are UCSD-only access should also have Public metadata-only access
- License\permission\Permission\type = "localDisplay"
- OtherRights\permission\Permission\type = "localDisplay"

### Expected behavior

Update access controls for downloading of both mp3, mp4, others?

UCSD IP users should see the content, public users should see the "Restricted content" image instead of content

mcritchlow:
I'm curious about this, as it related to work I'm doing with @jhriv  on serving files differently.

Is the issue that _any_ mp3 or mp4 can be downloaded by someone if they know the URL? Or that mp3's that are, say, curator only , can be downloaded by a non-curator? I'm curious what the context is that's currently incorrect.

lsitu:
@mcritchlow Good question! I think the context is that though we do not allow the audio/video viewer to have a download button but if someone know about the full url path, they could still be able to download the audio/video. It looks like it's something related to ticket [Remove download function for UCSD-only video objects](https://github.com/ucsdlib/damspas/issues/461), in which one could still be able to download the video even though the download button is removed.
@gamontoya Is it just related to those UCSD-only objects as what's specified in ticket https://github.com/ucsdlib/damspas/issues/461, or all public downloads of audios/videos?

mcritchlow:
@lsitu I see, thanks. I do see that there is no download button for mp3's, but there [is for mp4's](https://library.ucsd.edu/dc/object/bb2118519t). This seems odd to me, at the very least inconsistent.. Why can't people download mp3s when they're allowed to stream them, if they're derivatives? (question for @gamontoya )

I suppose my more concrete concern/issue, is that I want to make sure whatever access control checks necessary are performed before [`send_file` is called.](https://github.com/ucsdlib/damspas/blob/send-file/app/controllers/file_controller.rb#L65) Because per @jhriv and my testing so far, nginx and send_file will respect this and not allow direct queries to the URL's @gamontoya posted above, unless the access control checks allow them to.

gamontoya:
@GregReser @remerjohnson I'm adding Greg and Ryan to this discussion because they were in the meeting with me AND because they may have use cases that I may not be aware of.

GregReser:
[...]

PR #471, opened 2018-06-22
Remove download link for simple ucsd-only object

### LINKED ISSUE #461: Remove download function for UCSD-only video objects
### Descriptive summary

If the user is not logged as a curator, they should  should not be able to download videos for objects that are restricted to UCSD-only.  This should be true for all objects, components, and sub-components.
 
This should be true if the object contains either:
OtherRights\permission\Permission\type = "localDisplay"
License\permission\Permission\type = "localDisplay"

Example: https://librarytest.ucsd.edu/dc/object/bd66071852
image

A current ingest is dependent on this issue:
https://lib-jira.ucsd.edu:8443/browse/DI-224?filter=12684

VivianChu:
Fixes #461 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
Remove download function for UCSD-only video objects

References #461

### DIFF
--- a/app/views/dams_objects/_admin_download.html.erb
+++ b/app/views/dams_objects/_admin_download.html.erb
@@ -1,14 +1,12 @@
 <div id="etc-menu">
+  <% adl_glyph =  '<i class="glyphicon glyphicon-download-alt"></i>'.html_safe %>
+  <% adl_title = 'Download File' %>
+  <% adl_class = 'adl pull-right' %>
 
-	<% ucsd_local_display = (@document['otherRights_tesim'] != nil && @document['otherRights_tesim'].first.to_s.include?('localDisplay')) ? 'true' : 'false' %>
-	<% adl_glyph =  '<i class="glyphicon glyphicon-download-alt"></i>'.html_safe %>
-	<% adl_title = 'Download File' %>
-	<% adl_class = 'adl pull-right' %>
-
-	<% if can? :update, @document then %>
-	  <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
-	<% elsif ucsd_local_display == 'false' && !downloadDerivativePath.nil?%>
-	  <%= link_to adl_glyph, downloadDerivativePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
-	<% end %>
+  <% if can? :update, @document then %>
+    <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
+  <% elsif can_download?(@document) && !downloadDerivativePath.nil?%>
+    <%= link_to adl_glyph, downloadDerivativePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
+  <% end %>
 
 </div>
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,3 +46,9 @@ Metrics/LineLength:
 Style/SafeNavigation:
   Exclude:
     - 'lib/dams/controller_helper.rb'
+
+Metrics/CyclomaticComplexity:
+  Max: 8
+
+Metrics/PerceivedComplexity:
+  Max: 8
--- a/app/helpers/dams_objects_helper.rb
+++ b/app/helpers/dams_objects_helper.rb
@@ -2,7 +2,7 @@ require "base64"
 require "openssl"
 
 module DamsObjectsHelper
-
+  include Dams::ControllerHelper
   #---
     # Openurl, embedded metadata support,
     # Get metadata from solr document, and parsing each field and produce openURL key-encoded value query strings.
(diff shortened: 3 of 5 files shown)


app/views/dams_objects/_admin_download.html.erb
```diff
@@ -1,14 +1,19 @@
...
+  <% adl_class = 'adl pull-right' %>
 
-	<% if can? :update, @document then %>
-	  <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
-	<% elsif ucsd_local_display == 'false' && !downloadDerivativePath.nil?%>
-	  <%= link_to adl_glyph, downloadDerivativePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
-	<% end %>
+  <% if can? :update, @document then %>
+    <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:adl_class, title:adl_title %>
+  <% elsif local_otherRights == 'false' && local_license == 'false' && !downloadDerivativePath.nil?%>
```
mcritchlow: Could we potentially move this code in a helper method, something like `can_download?` (which takes in a solr document) or something? This would include the code above that checks the various Solr properties to set the values of `local_otherRights` and `local_license`

My thinking behind this is because while this code will do a good job preventing links from showing to users who shouldn't see them, I don't think this actually prevents those users from downloading the files directly through the file dowload controller.

So if [this line in the file controller](https://github.com/ucsdlib/damspas/blob/develop/app/controllers/file_controller.rb#L46) could be changed to something like:

`authorize!(:show, @solr_doc) && can_download?(@solr_doc)` 

Then we could enforce the logic in both places with the same code. Or something like that.

Curious what you think.

VivianChu: @mcritchlow - Sounds good.  I'll move the code in a helper method and update the file controller.  I think the issue of downloading the files directly is also mentioned in this issue https://github.com/ucsdlib/damspas/issues/467 and Longshou is working on it.

mcritchlow: @VivianChu - sounds great. Yeah I think this would hopefully take of both tickets in one PR potentially.

VivianChu:
@mcritchlow -  The code has been moved to the helper method.  However, if I update the file_controller to use authorize! :show, @solr_doc && can_download?(@solr_doc), then one test fails in feature/file_spec - https://github.com/ucsdlib/damspas/blob/feature/remove-download/spec/features/file_spec.rb#L69  I think maybe because of different return type - authorize! :show, @solr_doc returns a solr document while can_download? function returns a boolean.  When I use authorize! :show, @solr_doc if can_download?(@solr_doc), then the test passes. I try to test the file download for video object, & I can always be able to download it with no login or login as anonymous with all options in file_controller like with && can_download?(@solr_doc), if can_download?(@solr_doc), or without calling the can_download function.  This is the test I try and it always fails-   
scenario 'local user should not be able to download file for localDisplay license object' do
    visit file_path( @obj, '_1.mp4' )
    expect(page).to have_selector('h1', :text => 'Forbidden')
  end
Probably there are more things to do about file access control than just changing that line. 
Do you think if I should continue trying to fix the file download issue in this branch or let Longshou works on that for the other ticket?  Thanks

lsitu:
@mcritchlow /@VivianChu If file download is not straight forward with refactoring to the helper method, just leave it for ticket https://github.com/ucsdlib/damspas/issues/467 for me to work on then. Thanks.

mcritchlow:
@VivianChu @lsitu - I defer to you both on who gets the file controller working. I suspect if authorize method returns a solr document, then the call to `can_download?` could potentially be on the next line. Something like:
```
authorize! :show, @solr_doc
unless can_download?(@solr_doc)
  raise ActionController::RoutingError, "User is not allowed to download requested file: #{field}" 
end
```

VivianChu:
@mcritchlow - I use the above block of code and the test passes for local user.  However, it doesn't work correctly because now even if I use sign_in_developer to be a curator user, he/she can't download the file, he/she will see the forbidden text.

mcritchlow:
@VivianChu - ah i see, yeah that makes sense. we'd need to use a cancan check as a first conditional or something so that curators are allowed through regardless of the metadata checks in `can_download?` Or, you could have `can_download?`  eturn true immediately if the user is a curator. Something like that maybe? Whatever you think best.

lsitu:
@VivianChu Please feel free to leave it for ticket https://github.com/ucsdlib/damspas/issues/467 if you want, which is targeted the file download things.

VivianChu:
@mcritchlow @lsitu - I removed the can_download? check in file_controller so Longshou can work on his ticket about file download.  Could you review the changes?  Thanks

mcritchlow:
@VivianChu - Ok, got it. In that case, for the context of hiding links, as far as I can tell, this LGTM

lsitu:
👍 
@mcritchlow / @VivianChu Yeah, I agree that the context seems right with the PR and I see that the file download things may need  special check with CANCAN authorization, audio/videos context and thorough rspec testing, which is well under the context of ticket https://github.com/ucsdlib/damspas/issues/467.

Issue #466, opened 2018-06-20
Fix display behavior of metadata-only collections


gamontoya:
### Descriptive summary

- Center for Music Experiment
- Dept. of Music Concert Archive
- ECE Distinguished Lecturer Series
- Kelco Aerial Photograph Collection
- Scientific Voyages and Expedition Reports

Currently on staging, these collections are not behaving as expected. If I am on campus, I am seeing metadata-only behavior. 

### Expected behavior

- When on campus or using VPN (UCSD IP only), user should be able to view the metadata, see (images), or play files (audio/video), but should not be able to download (no download button either).

- When off campus, user will see the metadata only display with the banner on the object and generic thumbnails on grid view (no download button).

- When logged in as curators, user should be able to view the metadata, see (images), or play files (audio/video), and download the master files.

### Related work

#356

PR #473, opened 2018-06-27
Fixes #467 - Disable file downloading for localDisplay videos/audios and metadataDisplay objects.

### LINKED ISSUE #467: Fix access controls for downloading mp3 and mp4 files

lsitu:
Fixes #467

#### Local Checklist
- [x] Code style checked?
- [ ] QA-ed locally?

#### What does this PR do?
Disable file downloading for localDisplay videos/audios and metadataDisplay objects, while allowing campus users to download localDisplay images.

References #466, #467 

### DIFF
--- a/lib/dams/controller_helper.rb
+++ b/lib/dams/controller_helper.rb
@@ -914,12 +914,33 @@ module Dams
     #---
 
     def can_download?(document)
-      local_license = false
-      local_other_rights = document['otherRights_tesim'] && document['otherRights_tesim'].first.to_s.include?('localDisplay') ? true : false
-      if document['resource_type_tesim'] && document['resource_type_tesim'].first.to_s.include?('video')
-        local_license = document['license_tesim'] && document['license_tesim'].first.to_s.include?('localDisplay') ? true : false
+      license = document['license_tesim']
+      other_rights = document['otherRights_tesim']
+      return false if streaming_media_type?(document['resource_type_tesim']) && (local_display?(license) || local_display?(other_rights))
+      !metadata_only_display?(license) || !metadata_only_display?(other_rights)
+    end
+
+    def streaming_media_type?(resource_types)
+      resource_types.each do |t|
+        return true if t.include?('video') || t.include?('sound')
       end
-      !local_other_rights && !local_license
+      false
+    end
+
+    def local_display?(data)
+      return false unless data
+      data.each do |d|
+        return true if d.include?('localDisplay')
+      end
+      false
+    end
+
+    def metadata_only_display?(data)
+      return false unless data
+      data.each do |d|
+        return true if d.include?('metadataDisplay')
+      end
+      false
     end
   end
 end
--- a/lib/dams/controller_helper.rb
+++ b/lib/dams/controller_helper.rb
@@ -898,49 +898,45 @@ module Dams
       end
     end
 
-  def metadata_display?(data)
-    result = false
-    return result unless data
-    data.each do |datum|
-      result = true if datum.include?('localDisplay') || datum.include?('metadataDisplay')
+    def metadata_display?(data)
+      result = false
+      return result unless data
+      data.each do |datum|
+        result = true if datum.include?('localDisplay') || datum.include?('metadataDisplay')
+      end
+      result
     end
-    result
-  end
 
     #---
     # Check to see if an object allows file download
     #
     # @return Boolean value
     #---
+    def can_download?(solr_doc, fileid = nil)
+      permissions = (Array(solr_doc['license_tesim']) + Array(solr_doc['otherRights_tesim'])).flatten.compact
+      return false if streaming_media_type?(solr_doc['resource_type_tesim']) && local_display?(permissions)
+      metadata_only_display = metadata_only_display?(permissions)
+      return !restricted_file?(fileid) if fileid && metadata_only_display
+      !metadata_only_display
+    end
 
-    def can_download?(document)
-      license = document['license_tesim']
-      other_rights = document['otherRights_tesim']
-      return false if streaming_media_type?(document['resource_type_tesim']) && (local_display?(license) || local_display?(other_rights))
-      !metadata_only_display?(license) || !metadata_only_display?(other_rights)
+    def restricted_file?(fileid)
+      fileid && fileid.include?('_2.jpg')
     end
 
     def streaming_media_type?(resource_types)
-      resource_types.each do |t|
-        return true if t.include?('video') || t.include?('sound')
-      end
-      false
+      return false if resource_types.blank?
+      resource_types.any? { |t| t.include?('video') || t.include?('sound') }
     end
 
     def local_display?(data)
-      return false unless data
-      data.each do |d|
-        return true if d.include?('localDisplay')
-      end
-      false
+      return false if data.blank?
+      data.any? { |t| t.include?('localDisplay') }
     end
 
     def metadata_only_display?(data)
-      return false unless data
-      data.each do |d|
-        return true if d.include?('metadataDisplay')
-      end
-      false
+      return false if data.blank?
+      data.any? { |t| t.include?('metadataDisplay') }
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/dams/controller_helper.rb
```diff
@@ -914,12 +914,33 @@ def metadata_display?(data)
...
-      if document['resource_type_tesim'] && document['resource_type_tesim'].first.to_s.include?('video')
-        local_license = document['license_tesim'] && document['license_tesim'].first.to_s.include?('localDisplay') ? true : false
+      license = document['license_tesim']
+      other_rights = document['otherRights_tesim']
+      return false if streaming_media_type?(document['resource_type_tesim']) && (local_display?(license) || local_display?(other_rights))
+      !metadata_only_display?(license) || !metadata_only_display?(other_rights)
+    end
+
+    def streaming_media_type?(resource_types)
+      resource_types.each do |t|
```
mcritchlow: For these support methods you might be able to use the `.any?` Enumerable method to avoid the explicit returns. Something like:

```
def streaming_media_type?(resource_types)
  resource_types.any? { |t| t.include?('video') || t.include?('sound') }
end
```

spec/controllers/file_controller_spec.rb
```diff
@@ -0,0 +1,516 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
mcritchlow: This is a great test setup :clap: I'll have to tweak this once we fold in the `send_file` code, but I already have a pattern in place for that. And these are an excellent set of comprehensive tests.

lib/dams/controller_helper.rb
```diff
@@ -914,12 +914,33 @@ def metadata_display?(data)
     #---
 
     def can_download?(document)
-      local_license = false
-      local_other_rights = document['otherRights_tesim'] && document['otherRights_tesim'].first.to_s.include?('localDisplay') ? true : false
-      if document['resource_type_tesim'] && document['resource_type_tesim'].first.to_s.include?('video')
-        local_license = document['license_tesim'] && document['license_tesim'].first.to_s.include?('localDisplay') ? true : false
+      license = document['license_tesim']
+      other_rights = document['otherRights_tesim']
+      return false if streaming_media_type?(document['resource_type_tesim']) && (local_display?(license) || local_display?(other_rights))
```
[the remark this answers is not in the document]
lsitu: Sound good. I'll refactor it for the conditional combination, and I see it more complicated with those UCSD IP/metadata-only collections that won't allow local download but need VPN/campus view.

lsitu:
@mcritchlow Good point! I was thinking about it but it seem like we may need to add the file_id parameter to the can_download? method to support the specs in #466 that need those UCSD IP/metadata collections to be viewable from campus or through VPN (UCSD IP restriction).
In this case, we may need to disallow the downloading of files like _2.jpg, as well as other _1.* service files. So it seem like we should keep it as a method. What do you think? Do you think if we make a simple check with the filename will be enough, or need to look up and parse the data for all the files o check the file use property, which is not so efficient for a match?

mcritchlow:
@lsitu - Yeah i suspect it might be best to just leave your existing implementation as proposed in this PR as-is, because there's enough complexity in this that it's probably not worth it. it was just a thought that occurred to me :) hopefully next time we do this, in dams5, we'll have more rational metadata for making this decision. we'll see..

lsitu:
@mcritchlow Yep. It sounds good. I was thinking whether we should at least filter those _2.jpg from downloading or not. Thanks.

VivianChu:
👍

lib/dams/controller_helper.rb
```diff
@@ -898,28 +898,45 @@ def referrer_controller( request )
...
 
-    def can_download?(document)
-      local_license = false
-      local_other_rights = document['otherRights_tesim'] && document['otherRights_tesim'].first.to_s.include?('localDisplay') ? true : false
-      if document['resource_type_tesim'] && document['resource_type_tesim'].first.to_s.include?('video')
-        local_license = document['license_tesim'] && document['license_tesim'].first.to_s.include?('localDisplay') ? true : false
-      end
-      !local_other_rights && !local_license
+    def restricted_file?(fileid)
+      fileid && fileid.include?('_2.jpg')
```
mcritchlow: What is specific to an `_2.jpg` being a restricted file? I don't think I understand the meaning behind this one.

lib/dams/controller_helper.rb
```diff
@@ -898,28 +898,45 @@ def referrer_controller( request )
       end
     end
 
-  def metadata_display?(data)
-    result = false
-    return result unless data
-    data.each do |datum|
-      result = true if datum.include?('localDisplay') || datum.include?('metadataDisplay')
+    def metadata_display?(data)
```
mcritchlow: Since we already have this `metadata_display?` method, and it appears to be used in the catalog_helper and dams_resource_controller, do we need the `local_display?` and `metadata_only_display?` methods? Could we use this instead? This one seems to just check both (though it could be rewritten to use `any?` also like the others. Just a thought.

lsitu:
@mcritchlow Good question! 
To support UCSD IP/metadata-only collections with camps/VPN access behavior, we need to disable the download button and the downloading of the files, but we also need the objects to be viewable, which requires file download for the object viewers. Those _2.jpeg full screen resolution images are linking to the download button so I just restrict them from downloading at this time. I think we can add any other restricted file types from downloading when necessary. Just let me know what we want to start with and I think we can adjust it now or later. Thanks.

lib/dams/controller_helper.rb
lsitu: @mcritchlow I think we may need to a separate method for metadata_only_display? to determine the different behavior with localDisplay and metadataDisplay objects. For localDisplay objects,  they allow downloading of files (video/audio excluded) while metadataDisplay won't allow file download for any files.

mcritchlow: @lsitu - Ok, thanks for clarifying. Perhaps I misunderstood what Greg said. I thought he said if either were filled in then one could not download derivatives. But yeah if there are distinctions then I guess that's another issue. Poorly named properties it turns out, because they don't actually describe what they do :)

lsitu:
Yep, good questions again. I think "metadata only" is only for public users, things become complicated/confusing for campus /VPN (UCSD IP restricted) users. The issue is that why the download button need to be removed from the viewer for those UCSD IP/metadata-only collections. And as you noted, does those images link to the download button a matter? If the answer is no, then removing of the download button seems to be a question.

lib/dams/controller_helper.rb
lsitu: @mcritchlow Yes, it's confusing to me too. So I asked Greg to clarify the localDisplay thing earlier since the access control issue is always sensitive to us. Do you want to rename it?

mcritchlow: @lsitu - I'm not sure. Honestly, all your tests pass and I think it's a pretty comprehensive test suite. I don't want to slow this down more than necessary. I just think the data model implications and application to these access control scenarios is unsustainable, very fragile at the least. But that's hopefully an issue we can address in the next system. Perhaps unless @hweng or @VivianChu have any other concerns, I'll just merge this and we can move on? (Thanks for all your awesome work on this complicated ticket)

lsitu: @mcritchlow Could you hold on? I'll refactor it a little bit. Thanks.

lsitu:
@mcritchlow I've refactored the metadata_display? codes to use it to include audio/video download controlling for the UCSD IP/metadata-only collections with the additional tests added.

mcritchlow:
LGTM

hweng:
:+1:

PR #494, opened 2018-07-18
Add different access text for mixed objects metadata-only collection

### LINKED ISSUE #483: Change access text for mixed collection
### Descriptive summary

Display different access view type like "Some items restricted" for mixed collections.

### Expected behavior

It will display "Some items restricted" access type.

### Actual behavior

Currently, it displays "Restricted View" access type.

### Related work
#466

VivianChu:
Fixes #483 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
 
It will display "Some items restricted" access type for collection with mixed objects.

References #483 

#### Screenshots

mixedcollection

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -4,6 +4,7 @@ require 'rsolr'
 
 class CatalogController < ApplicationController  
   include Blacklight::Catalog
+  include Dams::ControllerHelper
   # Extend Blacklight::Catalog with Hydra behaviors (primarily editing).
   include Hydra::Controller::ControllerBehavior
   
@@ -250,7 +251,6 @@ class CatalogController < ApplicationController
 
   # get search results from the solr index
   def index
-    @metadata_colls = metadata_only_collections
     if params['q'] != nil
       params['q'].gsub!('""','')
       single_quote_count = params['q'].to_s.count('"') 
@@ -271,6 +271,7 @@ class CatalogController < ApplicationController
     end
 
     (@response, @document_list) = get_search_results
+    @metadata_colls = metadata_only_collections(params)
 
     @filters = params[:f] || []
     
@@ -283,7 +284,6 @@ class CatalogController < ApplicationController
   end
 
   def collection_search
-    @metadata_colls = metadata_only_collections
     # if we already have the parameters set below, then redirect to /search
     # this allows removing Collections limit, etc.
     #if (params[:sort] || (params[:fq] && params[:fq].to_s.include?('{!join')) )
@@ -316,15 +316,18 @@ class CatalogController < ApplicationController
     search[:fq] = params[:fq] if params[:fq]
     search[:total] = @response.response['numFound']
     session[:search] = search
+    @metadata_colls = metadata_only_collections(params)
   end
 
-  def metadata_only_collections
+  def metadata_only_collections(params)
     meta_colls = []
-    other_rights_fquery = '(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay)'
-    params = { fq: "{!join from=collections_tesim to=id}#{other_rights_fquery}", q: 'type_tesim:Collection', rows: 300 }
-    response = raw_solr(params)
+    solr_params = {}
+    solr_params[:fq] = ["{!join from=collections_tesim to=id}#{metadata_only_fquery}"]
+    response = raw_solr(solr_params.merge(params))
     response.docs.each do |doc|
-      meta_colls << doc['id_t']
+      mix_obj = mix_objects?(doc['id_t'])
+      val = mix_obj ? "#{doc['id_t']}#{mix_obj}" : doc['id_t']
+      meta_colls << val
     end
     meta_colls
   end
@@ -340,12 +343,7 @@ class CatalogController < ApplicationController
       logger.warn "Error looking up unit name: #{e}"
     end
   end
-  def raw_solr( params={} )
-    solr_path = blacklight_config.solr_path
-    res = blacklight_solr.send_and_receive(solr_path, :params => params)
-    solr_response = Blacklight::SolrResponse.new(force_to_utf8(res), params)
-    solr_response
-  end
+
   def solrdoc
     @obj = ActiveFedora::Base.find(params[:id], :cast=>true)
     xml = RSolr::Xml::Generator.new
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -325,7 +325,7 @@ class CatalogController < ApplicationController
     solr_params[:fq] = ["{!join from=collections_tesim to=id}#{metadata_only_fquery}"]
     response = raw_solr(solr_params.merge(params))
     response.docs.each do |doc|
-      mix_obj = mix_objects?(doc['id_t'])
+      mix_obj = mix_objects?(doc['id_t'], metadata_obj_count(doc['id_t']))
       val = mix_obj ? "#{doc['id_t']}#{mix_obj}" : doc['id_t']
       meta_colls << val
     end
--- a/app/controllers/dams_resource_controller.rb
+++ b/app/controllers/dams_resource_controller.rb
@@ -42,8 +42,8 @@ class DamsResourceController < ApplicationController
     if models.include?("DamsAssembledCollection") || models.include?("DamsProvenanceCollection") || models.include?("DamsProvenanceCollectionPart") 
         collection_solr_params = build_params("collections_tesim:#{params[:id]}", metadata_only_fquery)
         collection_response = raw_solr(collection_solr_params)
-        @metadata_only = !collection_response.response['numFound'].zero?
-        @mix_objects = mix_objects?(params[:id], collection_response.response['numFound']) if @metadata_only
+        @metadata_only = collection_response.response['numFound'].to_i > 0
+        @mix_objects = mix_objects?(params[:id], collection_response.response['numFound'].to_i)
         
         facet_collection_params = { :f=>{"collection_sim"=>"#{@document['title_tesim'].first.to_s}"}, :id=>params[:id], :rows => 0 }
         apply_gated_discovery( facet_collection_params, nil )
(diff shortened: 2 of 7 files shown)


lib/dams/controller_helper.rb
```diff
@@ -903,6 +903,37 @@ def metadata_display?(data)
...
+
+    def build_params(query, fquery = nil)
+      { q: query, fq: fquery, rows: 1 }
+    end
+
+    def metadata_only_fquery
+      '(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay OR license_tesim:localDisplay)'
+    end
+
+    def raw_solr(params = {})
```
lsitu: It seems like it's the same method as  https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR345, which need cleanup. But I am not sure why we need it since blacklight already provides a Solr helper function [SolrHelper:get_search_results](https://www.rubydoc.info/gems/blacklight/5.8.2/Blacklight%2FSolrHelper:get_search_results) to do it thought.

app/controllers/catalog_controller.rb
```diff
@@ -317,14 +318,15 @@ def collection_search
...
+  
   def metadata_only_collections
     meta_colls = []
-    other_rights_fquery = '(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay)'
-    params = { fq: "{!join from=collections_tesim to=id}#{other_rights_fquery}", q: 'type_tesim:Collection', rows: 300 }
+    params = { fq: "{!join from=collections_tesim to=id}#{metadata_only_fquery}", q: 'type_tesim:Collection', rows: 300 }
     response = raw_solr(params)
     response.docs.each do |doc|
-      meta_colls << doc['id_t']
+      mix_obj = mix_objects?(doc['id_t'])
```
lsitu: Will the recurring checks with Solr queries for all metadata_only collection slowdown the page loading or not? Is it possible to do it in the page level?

lib/dams/controller_helper.rb
VivianChu: @lsitu - I use the get_search_results before and it always returns different numFound.  https://github.com/ucsdlib/damspas/pull/494/files#diff-c0623f718c6b24f20f5334d0d4eaf8d5L45
I see this existing function like lookup_unit_name also uses raw_solr()
https://github.com/ucsdlib/damspas/blob/develop/app/controllers/catalog_controller.rb#L336
I use it and can see that it returns the correct numFound for my function.

app/controllers/catalog_controller.rb
VivianChu: @lsitu - I think it will slow down the page loading but not much because we don't have that much metadata_only_collection.  Do you have any suggestion of how to do it in the page level?  Thanks

lib/dams/controller_helper.rb
lsitu: Yep, it seems like a long history for that method. Just curios about it. But I think get_search_results should work as expected for the counts: https://github.com/ucsdlib/damspas/blob/develop/app/controllers/dams_units_controller.rb#L27.

app/controllers/catalog_controller.rb
lsitu: What causing my attention is that if we just have 10 such collections, we need to run 20 SOLR queries for each page loading, even with pages that are all objects. That could be a huge burden for our servers.
Maybe we can just check that with the records in search results in each page to see whether it contains metadata-only collections, and then just check for those metadata-only collections instead?
https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR311
https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR274

lib/dams/controller_helper.rb
lsitu: @VivianChu I did some tests with your commit to use get_search_results and I do see the issue that it failed to retrieve the objects in catalog controller with a simple query to retrieve all the records in a collection like {:q=>"collections_tesim:xx00063001", :rows=>1}.
Maybe we just clean up the duplicate definition in https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR345 and use raw_solr to avoid that annoying issue?

VivianChu: @lsitu - Sounds good.  The code has been updated to remove the duplicate definition and use raw_solr. Thanks

lsitu:
@VivianChu  👍It looks good to me. Thanks for the nice improvement!

app/helpers/catalog_helper.rb
```diff
@@ -96,8 +96,9 @@ def display_access_control_level(document, metadata_colls)
     access_group = document['read_access_group_ssim'] # "public" > "local" > "dams-curator" == "dams-rci" == default
     view_access = 'Curator Only'
     if access_group
-
-      if metadata_colls.include?(document['id']) || metadata_display?(rights_data(document))
+      if metadata_colls.include?("#{document['id']}true")
```
mcritchlow: What is the `true` part here? Is that part of the query response, or?

app/controllers/dams_resource_controller.rb
```diff
@@ -40,11 +40,11 @@ def show
...
     models = @document["active_fedora_model_ssi"]
     if models.include?("DamsAssembledCollection") || models.include?("DamsProvenanceCollection") || models.include?("DamsProvenanceCollectionPart") 
-        other_rights_fquery = "(otherRights_tesim:localDisplay OR otherRights_tesim:metadataDisplay)"
-        collection_solr_params = { :q => "collections_tesim:#{params[:id]}", :fq => other_rights_fquery, :rows => 1 }
-        collection_response, collection_documents = get_search_results(collection_solr_params)
+        collection_solr_params = build_params("collections_tesim:#{params[:id]}", metadata_only_fquery)
+        collection_response = raw_solr(collection_solr_params)
         @metadata_only = !collection_response.response['numFound'].zero?
-
+        @mix_objects = mix_objects?(params[:id], collection_response.response['numFound']) if @metadata_only
```
mcritchlow: If `collection_response.response['numFound']` is `0` or `nil`, do we still need to call this method? It seems, after a first read, that is that is the situation. Which makes me think that `mix_objects?` is going to run the exact same query again within its method, and would therefore get the same result.

The first seems to be a few lines up w/ `collection_solr_params = build_params("collections_tesim:#{params[:id]}", metadata_only_fquery)`

The second is in the `metadata_obj_count` helper method that gets called in `mix_objects?`.

 Apologies if I read this wrong, and if so please ignore. But if `numFound` being `0` or `nil` means `mix_objects?` should ultimately return `false`, we could possibly save a query?

VivianChu: @mcritchlow - I think the method mix_objects? will return false when numFound is 0 or nil for this case.  The method doesn't run the same query twice. It will only runs that query when the other place doesn't pass in that numFound like for browse by collection place  https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR328  I can add extra line above to make @mix_objects = false by default https://github.com/ucsdlib/damspas/pull/494/files#diff-c0623f718c6b24f20f5334d0d4eaf8d5R46 What do you think?

Could you clarify more about how to save a query?  That query is only execute for browse by collection or search a collection page.  For view a single collection page, that query won't get execute twice.

Thanks

app/helpers/catalog_helper.rb
VivianChu: It's just a tag there to indicate that's a mixed object collection build from this line
https://github.com/ucsdlib/damspas/pull/494/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR329

app/controllers/dams_resource_controller.rb
mcritchlow: @VivianChu - Yeah, I may likely just be misunderstanding. What I meant by save a query, was I see the first query happening here https://github.com/ucsdlib/damspas/blob/b1d806622c38c1ee6502041df03d35c61d564986/app/controllers/dams_resource_controller.rb#L43-L44

And the response `numFound` being used to pass into the `mix_objects?` method:
https://github.com/ucsdlib/damspas/blob/b1d806622c38c1ee6502041df03d35c61d564986/app/controllers/dams_resource_controller.rb#L46

 and, it seems, to call `metadata_object_count` if the response is empty https://github.com/ucsdlib/damspas/blob/b1d806622c38c1ee6502041df03d35c61d564986/lib/dams/controller_helper.rb#L932 and then the query being run there looks to me to be the same as the original https://github.com/ucsdlib/damspas/blob/b1d806622c38c1ee6502041df03d35c61d564986/lib/dams/controller_helper.rb#L912

So the 'save a query' idea would be not running that second query if the original `numFound` is `nil`, since it seemed to then do the same thing and return `false` if there's no count. But again, i may just be reading this wrong :)

app/helpers/catalog_helper.rb
mcritchlow: Ah, I see. Would it be worth adding a comment or something for that? Now that you point it out that makes sense, but it wasn't obvious to me initially that was part of the solution. Especially since `true` is a keyword, so my brain defaults to that i guess.

app/controllers/dams_resource_controller.rb
lsitu: @VivianChu Maybe we can change the line 
https://github.com/ucsdlib/damspas/pull/494#discussion-diff-204544243R45 to something like `@metadata_only = collection_response.response['numFound'] > 0`, which is more easy to read, and this could be an error to avoid the nil case if it does exist?

mcritchlow: Oh, I'm sorry. I missed this https://github.com/ucsdlib/damspas/pull/494/files#diff-c0623f718c6b24f20f5334d0d4eaf8d5R45 where you explicitly check for this. 

Perhaps what I wonder about then is if the default `nil` value and conditional to check for it are needed in `mix_objects?`? Is there any way then for that method to be called with there _aren't_ any results?

lsitu: @mcritchlow It looks like solr will always return numFound="0" if no results are found for a query. But if the server is mis-behavior, then it should be the error handling issue.

mcritchlow: @lsitu - yeah, you're right. I think that assumption is perfectly reasonable. I suppose we _could_ tack on a `.to_i` to the end. But you're right I don't think that's needed at all.

It's the checking again in the `mix_objects?` method (see my edit) and the resulting second query, that I'm less clear on. Apologies @VivianChu  that I missed that guard check with `@metadata_only`

PR #499, opened 2018-07-24
Remove host, username, password from db defaults

mcritchlow:
#### Local Checklist
- [x] QA-ed locally?

#### What does this PR do?
Removes the `ENV` fetching of environment variables from the `default` database environment block, and ensure that each environment handles these individually.

These default `DEV` ENV variables won't exist in all environments, and thus the fetch
calls will fail. Remove, and allow individual environments to specify as
needed.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -6,9 +6,6 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  host: <%= ENV.fetch('APPS_DHH_DEV_DB') %>
-  username: <%= ENV.fetch('APPS_DHH_DEV_DB_USER') %>
-  password: <%= ENV.fetch('APPS_DHH_DEV_DB_PASS') %>
   pool: 5
 variables:
   statement_timeout: 5000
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -6,16 +6,14 @@
...
-  username: <%= ENV.fetch('APPS_DHH_DEV_DB_USER') %>
-  password: <%= ENV.fetch('APPS_DHH_DEV_DB_PASS') %>
   pool: 5
 variables:
   statement_timeout: 5000
 
 development:
   <<: *default
   database: damspas_development
+  host: <%= ENV.fetch('APPS_DHH_DEV_DB') %>
```
lsitu: Do we still want a default value like 'dams' here? Just curios.

lsitu:
👍

config/database.yml
mcritchlow: Yeah, you know we definitely could. Each docker-compose file and the circleci config set these values. So, in theory, we shouldn't be in the situation where this value isn't set in the `development` or `test` environment context. That said, we could default to `localhost`, which seems kinda safe?

lsitu: Yep, it sounds good to me.

mcritchlow: Ok I'll do that then, thanks.

hweng:
:+1:

PR #503, opened 2018-08-02
Fix display behavior of metadata-only collections

### LINKED ISSUE #466: Fix display behavior of metadata-only collections

VivianChu:
Fixes #466 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
Fix restricted thumbnail issue for search results page.
Fix pdf file download issue for Scientific Voyages and Expedition Reports collection.
Add hide access label function for local user in search/browse by collection page, single object and collection view for metadata-only collection.

References #466 

#### Deployment Instructions
It has been deployed to staging and tested by Gabriela, Longshou and me.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -321,8 +321,7 @@ class CatalogController < ApplicationController
 
   def metadata_only_collections(params)
     meta_colls = []
-    solr_params = {}
-    solr_params[:fq] = ["{!join from=collections_tesim to=id}#{metadata_only_fquery}"]
+    solr_params = { q: "{!join from=collections_tesim to=id}#{metadata_only_fquery}", fq: 'type_tesim:Collection', rows: 300 }
     response = raw_solr(solr_params.merge(params))
     response.docs.each do |doc|
       mix_obj = mix_objects?(doc['id_t'], metadata_obj_count(doc['id_t']))
--- a/spec/features/catalog_spec.rb
+++ b/spec/features/catalog_spec.rb
@@ -613,7 +613,7 @@ feature 'Visitor wants to view object for public metadata-only collection in the
     expect(page).to have_content('Restricted View')
   end
 
-  scenario 'local user should not see the grey generic thumbnail and still see restricted access info' do
+  scenario 'local user should not see the grey generic thumbnail and restricted access label' do
     sign_in_anonymous '132.239.0.3'
     visit catalog_index_path({:q => 'xx909090zz'})
     expect(page).to_not have_css('img.dams-search-thumbnail[src="https://library.ucsd.edu/assets/dams/site/thumb-restricted.png"]')
--- a/spec/features/catalog_spec.rb
+++ b/spec/features/catalog_spec.rb
@@ -518,11 +518,25 @@ feature 'Visitor wants to view object of metadata data-only visibility collectio
     @localOnlyCollection.delete
     @localObj.delete
   end
-  scenario 'rendering the grey generic thumbnail and restricted access info' do
-    sign_in_developer
+
+  scenario 'public user should see the grey generic thumbnail and restricted access info' do
     visit catalog_index_path({:q => 'xx909090zz'})
     expect(page).to have_css('img.dams-search-thumbnail[src="https://library.ucsd.edu/assets/dams/site/thumb-restricted.png"]')
     expect(page).to have_content('Restricted View')
+  end
+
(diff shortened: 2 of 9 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -321,8 +321,7 @@ def collection_search
 
   def metadata_only_collections(params)
     meta_colls = []
-    solr_params = {}
-    solr_params[:fq] = ["{!join from=collections_tesim to=id}#{metadata_only_fquery}"]
+    solr_params = { q: "{!join from=collections_tesim to=id}#{metadata_only_fquery}", fq: 'type_tesim:Collection', rows: 300 }
```
lsitu: Is there any side effects when merging the parameter :q with that :q param passing in hash params in the line#325? Maybe we can try to use something like 
`{ fq: ["{!join from=collections_tesim to=id}#{metadata_only_fquery}", 'type_tesim:Collection'], rows: 300 }`?

spec/features/catalog_spec.rb
```diff
@@ -586,17 +613,60 @@
     expect(page).to have_content('Restricted View')
   end
 
-  scenario 'local user should see the grey generic thumbnail and restricted access info' do
+  scenario 'local user should not see the grey generic thumbnail and still see restricted access info' do
```
lsitu: Could we change ` still see` to match it with the expected test result `expect(page).to_not have_content('Restricted View')`?

lsitu:
@VivianChu It looks good to me. Just a couple of minor questions. 👍

spec/features/catalog_spec.rb
VivianChu: @lsitu - I changed the expected test result but forgot to change the scenario.  I'll update it.  Thanks for catching it.

app/controllers/catalog_controller.rb
VivianChu: @lsitu - I don't know if there's any side effect.  I could try your suggestion and deploy to staging to test.  Last time, it has problem on staging with the old fq.

spec/features/catalog_spec.rb
VivianChu: @lsitu - I deployed the change with that suggested query to staging.  Other collections from #466 worked fine.  However, some mixed collection and some sample collections that Greg created didn't work.  For example - https://librarytest.ucsd.edu/dc/search?utf8=%E2%9C%93&q=visibility   As curator, I see access label = Restricted to UC San Diego use only, but if I click on that link I see access label = Some items restricted.  The old query also worked fast too.  I don't see any different in performance between the old and the new query.  Do you have any other suggestion that I can try?  Thanks

app/controllers/catalog_controller.rb
lsitu: @mcritchlow I'll be off line tomorrow and I don't want to block the PR. Could you give it a pass if you don't see any issues with the query? I am not sure how merging of two :q parameter works in https://github.com/ucsdlib/damspas/pull/503/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR325.

VivianChu: @lsitu - I reverted the change to my query and deployed to staging. The search results worked fine. Both collection view and search view access label matched. It’s hard to have a test for it since staging will return larger results then local test with only few records created. There are already some tests to catalog spec and dams-collection test for search result access label. Those tests passed for both of our queries. They just behaved differently on staging.

mcritchlow:
This LGTM. The test coverage looks comprehensive and takes into account the various (there are so many..) edge cases I've seen discussed on the ticket. Nice work @VivianChu :+1:

mcritchlow:
@hweng - Any concerns with me merging this from your perspective?

hweng:
@mcritchlow Would you like to squash all those commits, or do you want me to do it?

hweng:
:+1:

PR #511, opened 2018-08-06
Document loading db schema in docker dev/test environments

mcritchlow:
#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?
- [x] Documentation updated (if needed)?

#### What does this PR do?
- Documents the change to use `db:schema:load` instead of `db:migrate` when using Docker for dev/test work. This is not only faster, but also ensures the `db/schema.rb` file doesn't get updated, which should really only happen when we add new migrations.
- Documents removing docker volumes along with containers.
- Updates CircleCI config to use `db:schema:load` for consistency (and minor speed gain)

- Prevent accidental commits of modified `db/schema.rb` file. And speed.

- `docker/README.md`

#### Manual testing steps?
- Feel free to run this locally and confirm if you're up for it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ jobs:
           command: docker-compose -f docker/ci/docker-compose.yml up -d
       - run:
           name: Setup Database
-          command: bundle exec rake db:create db:migrate
+          command: bundle exec rake db:create db:schema:load
       - run:
           name: Run Test Suite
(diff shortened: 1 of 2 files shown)


hweng:
:+1:

VivianChu:
👍

mcritchlow:
@ucsdlib/developers - thanks for your patience. got circleci using the new db:schema:load command as well now.

lsitu:
👍

PR #532, opened 2018-09-19
Notch8 Dev Setup

orangewolf:
#### Local Checklist
- [x] Tests written and passing locally? (N/A)
- [x] Code style checked? (N/A)
- [x] QA-ed locally? 
- [x] Configuration updated (if needed)? (N/A)
Not sure if you want to update the docs to use the command snippet or not

#### What does this PR do?

added env variables for dev proxy, shortcut script to launch docker-compose commands, fixed node/npm package split issue

Getting set up to make sure N8 development goes smoothly. We use dory, one of a bunch of proxy servers that all use the VIRTUAL_PORT / VIRTUAL_HOST convention.

Check the config changes, rebuild the docker image and see if the new version builds for you.

#### Deployment Instructions

No real deployment

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,5 @@
 Hydra::Application.routes.draw do
 
-  namespace :aeon do
-    resources :queues
-  end
   resources :audits, :only => [:index, :show]
   resources :pages
   get '/p/:id', to: 'pages#view', :as => 'view_page'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,8 @@
 Hydra::Application.routes.draw do
 
+  namespace :aeon do
+    resources :queues
+  end
   resources :audits, :only => [:index, :show]
   resources :pages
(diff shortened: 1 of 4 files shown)


mcritchlow:
@orangewolf - this looks good to me. After our mtg today I remembered that I had intended to make an update to the dev compose file. So I'll do that after this in a PR and tag you on it so you can give me your thoughts. Will merge on CircleCI is happy

config/routes.rb
```diff
@@ -1,5 +1,8 @@
 Hydra::Application.routes.draw do
 
+  namespace :aeon do
```
mcritchlow: just checking, do you want this bit included in the PR now?

orangewolf: nope, that goes in a separate branch. got bit by letting `git commit -am` become a habit back when git was new and we didn't know any better ;-)
I've reverted the extra change

PR #533, opened 2018-09-19
docker: enable debugging with byebug and friends

mcritchlow:
Fixes #issuenumber

#### Local Checklist
- [x] QA-ed locally?
- [x] Configuration updated (if needed)?
- [x] Documentation updated (if needed)?

#### What does this PR do?
Adds support and documentation for use debuggers like byebug with the docker/compose development environment

This is still new territory for us, and I ran across this issue working on a different project last week. So I thought I would document and add the solution that I found here.

See [`docker/README.md`](https://github.com/ucsdlib/damspas/blob/c2aeaef13faaaa1aa4137122dd119c89e0b63db5/docker/README.md#debugging-w-byebug) for how this would be used in practice

#### Manual testing steps?
- Run a fresh build using the `docker/dev/docker-compose.yml` file.
- Add a `byebug` statement somewhere. Controller, wherever.
- Run the app and get it to trigger the breakpoint
- Follow the instructions for using `docker attach <container>` and try interacting with byebug

@ucsdlib/developers and @orangewolf - please review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,7 @@ gem 'coveralls', '0.8.21', require: false
 gem 'rubyzip', '1.2.2'
 
 group :development, :test do
+  gem 'byebug'
   gem 'sass-rails', '~> 5.0.5' # locked
   gem 'coffee-rails', '~> 4.2.1'
   gem 'pry', '~> 0.10.4'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -91,6 +91,7 @@ GEM
       sass (~> 3.2)
     bower-rails (0.11.0)
     builder (3.2.3)
+    byebug (10.0.2)
     cancan (1.6.10)
     capistrano (3.8.1)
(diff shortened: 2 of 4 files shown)


hweng:
:+1:

mcritchlow:
@lsitu - that's a good point. this is more on the assumption that one might use byebug, or another debugger that requires a terminal to interact with, like pry. So the error you're seeing makes sense, byebug isn't actually in our Gemfile in the test/dev group. I can add it as part of this PR if folks would like, but I was admittedly writing this as something that _could_ support it. Thoughts?

lsitu:
@mcritchlow I am fine to have byebug gem added to test/dev group to make it works. But I think we can just update the documentation for it.

mcritchlow:
@lsitu - Ok sounds good. I'll add it to the Gemfile in a separate commit. Thanks for the feedback!

PR #627, opened 2019-05-03
Presist User Level Permissions

orangewolf:
Added initialize and save hooks to set the user level permissions based on the work authorizations table. This works around the fact that user level permissions are currently not persisted in the dams system

@mcritchlow the big thing I think to focus on here is to check if DamsObject is the right spot for this work? Is that the common denominator object?

### DIFF
--- a/app/models/dams_object.rb
+++ b/app/models/dams_object.rb
@@ -1,10 +1,14 @@
 class DamsObject < ActiveFedora::Base
   include Hydra::AccessControls::Permissions
   include Hydra::ModelMethods
+
+  after_initialize :setup_read_users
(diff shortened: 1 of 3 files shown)


mcritchlow:
@orangewolf yeah, I think this is right. I don't believe VRR grants permissions to anything other than objects (no collection-level stuff, for example), so I think this should do the trick.

@lsitu would you mind taking a look too when you have a moment?

mcritchlow:
I'm voting/merging in favor of testing this so we can make progress on having local folks test this out, on the hope that we're sorted out w/ persistence/solr indexing.

PR #630, opened 2019-05-09
Add VRR to damspas

mcritchlow:
This is all the work from Notch8 to implement VRR in damspas.

Additionally, it replaces NewRelic with Honecomb.

New things of particular note are:
- Database migrations to support Work Authorizations
- Custom auth using email unique IDs that map to the work authorizations
- Whenever gem for the hourly cron tasks

The unbelievably massive number of commits is quite unfortunate. That said, the majority of the new code is in `lib/processors`, the model for work authorization in `app/models/work_authorization` and the AEON MVC's in `app/{models|views|controllers}/aeon`

This has been running on staging for some months now, which is the best 'real' testing we can get.

@ucsdlib/developers - please review and let me know if you have any concerns. At this point, it's probably best to just get this in, and make adjustments going forward if/when needed.


VivianChu:
@mcritchlow - I approved the review but I couldn't merge because of the conflicts.  Could you take a look at the conflict?  Thanks

lsitu:
👍

mcritchlow:
@VivianChu - It's merged, thanks all!

PR #659, opened 2019-06-19
Add case-insensitive sorting for metadata fields; capitalize first character

### LINKED ISSUE #633: Multiple value fields not sorting alphabetically
From dams-metadata created by [GregReser](https://github.com/GregReser): ucsdlib/dams-metadata#43  

### Description
It appears that there is no sort logic for many fields in the DAMS 4 UI. Several project managers and data providers have requested that this be changed to default to alphabetic sort, if a specific order is not specified.

### Example of random sort:
image

### Request
For all fields that allow multiple values, if the display order is not specified, display alphabetically

mcritchlow:
Fixes #633 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
- Creates an `upcase_first` method this is backported from Rails 5, which supports the request from @ucsdlib/domm that the lists are normalized to capitalization. See screenshots below for an example of the difference
- Update view partials to sort fields case-insentively
- Update view partials to use `upcase_first` when rendering the text for the field
- Fix a few tests that broke with this new requirement of capitalization

References #633 

#### Screenshots
Before `upcase_first` (sorted, but leaving metadata values as-is)

pic-selected-190619-1010-06

With `upcase_first` and sorted case-insensitively

pic-selected-190619-1012-09

### DIFF
--- a/app/views/shared/fields/_relationships_data.html.erb
+++ b/app/views/shared/fields/_relationships_data.html.erb
@@ -4,8 +4,8 @@
 %>
 
 	<ul class="unstyled glyphicon-list glyphicon-list-tag">
-		<% value.sort.each do |name| %>
-			<li><%= link_to name, facet_uri(name, 'creator_sim') %></li>
+		<% value.each do |name| %>
+			<li><%= link_to name.upcase_first, facet_uri(name, 'creator_sim') %></li>
 		<% end %>
 	</ul>
 
--- a/app/views/shared/fields/_anatomy.html.erb
+++ b/app/views/shared/fields/_anatomy.html.erb
@@ -18,8 +18,8 @@
 %>
 
 	  <ul class="unstyled glyphicon-list glyphicon-list-tag">
-	    <% fieldData.each do |topic| %>
-	      <li><%= link_to topic, facet_uri(topic, 'subject_anatomy_sim') %></li>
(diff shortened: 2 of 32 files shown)


mcritchlow:
@lsitu @hweng - I heard from @GregReser and @arwenhutt that this looks good to them, so please feel free to review the PR at your convenience, thanks!

app/views/shared/fields/_relationships_data.html.erb
```diff
@@ -4,9 +4,9 @@
 %>
 
 	<ul class="unstyled glyphicon-list glyphicon-list-tag">
-		<% value.sort.each do |name| %>
-			<li><%= link_to name, facet_uri(name, 'creator_sim') %></li>
+		<% value.each do |name| %>
```
lsitu: Are there any reasons for not sorting names?

mcritchlow: @lsitu - oh nice, i think there was an issue w/ relationship names and not having a sort_by method, perhaps the class was different or something. I'll double check.

mcritchlow: @lsitu - thanks for catching that :) I must have just missed this one, and am thinking of a different file that I couldn't apply `sort_by` to. Just pushed an amended commit.

PR #663, opened 2019-06-19
RDCP: Sort by system modified date in browse by collection view

### LINKED ISSUE #644: Sort collections view by most recently added, by default
### Descriptive summary

Collections view for RDC (https://library.ucsd.edu/dc/rdcp/collections) should sort, by default, by most recently added collections, rather than the current alphabetic sort. 

### Rationale

This is a nearly ubiquitous practice in research data repositories, and emphasizes the most recently added collections rather than those beginning with "A". Alphabetic sort by title can remain an option under Sort.

### Note

- ONLY for RDCP collections.

mcritchlow:
Fixes #644 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
- Sorts rdcp browse by collection view by `system_modified` date descending, with `title` ascending as a secondary sort parameter
- Adds tests demonstrating this behavior for RDCP, but **not** for other units (DLDP)
- Adds `timecop` gem to easily support time-sensitive tests, which were relevant for the behavior in this PR
- Fixed a leaky test unit object that showed up in a [failed CI run](https://circleci.com/gh/ucsdlib/damspas/1511). This is not completely related, but seems problematic and worth fixing (I worry we have other examples of this..)

References #644 - for RDCP

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -2,12 +2,12 @@
 require 'blacklight/catalog'
 require 'rsolr'
 
-class CatalogController < ApplicationController  
+class CatalogController < ApplicationController
   include Blacklight::Catalog
   include Dams::ControllerHelper
   # Extend Blacklight::Catalog with Hydra behaviors (primarily editing).
   include Hydra::Controller::ControllerBehavior
-  
+
   # support boolean operators, even in basic search
   include BlacklightAdvancedSearch::ParseBasicQ
 
@@ -29,7 +29,7 @@ class CatalogController < ApplicationController
   # convert unit joins to normal facet queries (unless type=collection)
   CatalogController.solr_search_params_logic += [:transform_unit_scope]
 
- 
+
   def discovery_permissions
     @discovery_permissions ||= ["discover"]
   end
@@ -62,13 +62,13 @@ class CatalogController < ApplicationController
   end
 
   configure_blacklight do |config|
-  
-    config.default_solr_params = { 
+
+    config.default_solr_params = {
       :qt => 'search',
       :rows => 20,
       :qf => 'title_tesim^99 name_tesim^20 subject_tesim^10 scopeContentNote_tesim^6 all_fields_tesim fulltext_tesim id',
     }
-  
+
   config.advanced_search = {
       :form_solr_parameters => {
         "facet.field" => ["collection_sim", "object_type_sim", "unit_sim"],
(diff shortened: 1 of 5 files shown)


spec/features/collections_search.rb
```diff
@@ -0,0 +1,57 @@
...
+      @unit.delete
+      @oldest_collection.delete
+      @newest_collection.delete
+    end
+    it 'should sort by most recent collection, descending to oldest' do
+      visit dams_unit_collections_path('rdcp')
+      old_collection_sort_order = page.body.index('A Old Collection')
+      new_collection_sort_order = page.body.index('The New Collection')
+      expect(old_collection_sort_order).to be > new_collection_sort_order
+    end
```
lsitu: Could we add a test for user selected sorting rules like` sort by collection title` to see whether it works? It seems like there could be a gap but I may be wrong. See my other comment above.

app/controllers/catalog_controller.rb
```diff
@@ -307,6 +307,10 @@ def collection_search
 
     # sort by title
     params[:sort] = 'title_ssi asc' unless params[:sort]
+
+    # RDCP prefers their collections sorted by most recently created first
+    # Therefore, we're sorting by modified date rather than title
+    params[:sort] = 'system_modified_dtsi desc, title_ssi asc' if params[:id].eql? 'rdcp'
```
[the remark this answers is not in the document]
mcritchlow: My thinking was that this would be last in the chain of sorting checks. If I put it above, then it could potentially get overwritten later. But I may be misunderstanding you, can you help clarify with an example maybe?

spec/features/collections_search.rb
mcritchlow: The default sort I think is tested in the DLDP browse test I wrote below the RDCP one, which basically demonstrates the normal alphabetical sort. Can you take a look at that again and let me know if you think there's still a gap that needs testing? It's certainly possible :)

mcritchlow: @lsitu - Oh, I apologize. You're saying if a user selects different sorting rules after the page has initially loaded. That's a really great point, and possibly a regression. The current code might always force this sorting for RDCP. I understand what you're saying now, thanks for pointing this out. I'll think about how to adjust for this.

app/controllers/catalog_controller.rb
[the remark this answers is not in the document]
lsitu: @mcritchlow I think it's just related to your last comment in https://github.com/ucsdlib/damspas/pull/663#discussion_r295917873. The default `sort by title` will be set with the `unless params[:sort]` in https://github.com/ucsdlib/damspas/pull/663/files#diff-956e46b33fb5307990c4e1a4b5fd86ccR309 only when there are no sort order rules are selected.

mcritchlow:
@lsitu - Thank you again for catching the user-selected sorting regression! I've changed the code in catalog controller to only apply those defaults if no sorting option has been selected by the user. And I added a test to confirm that.

Let me know if you think this works.

hweng:
:+1:

PR #666, opened 2019-07-01
VRR - Disable download and embed buttons for authorized works view in…

### LINKED ISSUE #654: Disable download and embed buttons for authorized works view in VRR
### Descriptive summary

In the authorized user view of a VRR object, remove the download and embed buttons from the object view so authorized VRR users cannot download files or try to generate embed links that won't work due to the restricted nature of the materials.

### Related work

- https://github.com/ucsdlib/damsmanager/issues/332
- https://github.com/ucsdlib/damsmanager/issues/331
- https://github.com/ucsdlib/damsmanager/issues/320

hweng:
Fixes #654 

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
In the authorized user view of a VRR object, remove the download and embed buttons from the object view so authorized VRR users cannot download files or try to generate embed links

References #654

#### Manual testing steps?
Matt Peter and I have tested it on Staging that download and embed wasn't showing up on either the pdf or image objects I had in VRR queues.

### DIFF
--- a/app/views/dams_objects/_admin_download.html.erb
+++ b/app/views/dams_objects/_admin_download.html.erb
@@ -15,17 +15,20 @@
   end
 %>
 
-<div class="etc-menu">
 
-  <%= link_to embed_glyph, "#embedLink-#{component_id}", class:etc_menu_item_class, role:'button', :data => { :toggle => 'modal' }, title:'Embed' %>
+  <% if !(current_user && current_user.work_authorizations_count > 0) %>
+    <div class="etc-menu">
 
-  <% if can? :update, @document then %>
-    <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:etc_menu_item_class, title:'Download File' %>
-  <% elsif can?(:read, @document) && can_download?(@document) && !downloadDerivativePath.nil?%>
-    <%= link_to adl_glyph, downloadDerivativePath, :rel => 'nofollow', class:etc_menu_item_class, title:'Download File' %>
+      <%= link_to embed_glyph, "#embedLink-#{component_id}", class:etc_menu_item_class, role:'button', :data => { :toggle => 'modal' }, title:'Embed' %>
+
+      <% if can? :update, @document then %>
+        <%= link_to adl_glyph, downloadFilePath, :rel => 'nofollow', class:etc_menu_item_class, title:'Download File' %>
+      <% elsif can?(:read, @document) && can_download?(@document) && !downloadDerivativePath.nil?%>
+        <%= link_to adl_glyph, downloadDerivativePath, :rel => 'nofollow', class:etc_menu_item_class, title:'Download File' %>
+      <% end %>
+   </div>
   <% end %>
 
-</div>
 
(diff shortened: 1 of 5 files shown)


app/views/dams_objects/_admin_download.html.erb
```diff
@@ -15,17 +15,20 @@
   end
 %>
 
-<div class="etc-menu">
 
-  <%= link_to embed_glyph, "#embedLink-#{component_id}", class:etc_menu_item_class, role:'button', :data => { :toggle => 'modal' }, title:'Embed' %>
+  <% if !(current_user && current_user.work_authorizations_count > 0) %>
```
mcritchlow: This check is performed here, as well as in the `data_viewer` partial. I wonder if we could make this a little helper method, something like `vrr_user?` or something?

spec/features/dams_object_spec.rb
```diff
@@ -841,6 +841,7 @@
     @unit.delete
   end
   it "should show a 'View file' button and a 'Download file' button " do
+    sign_in_curator
```
mcritchlow: Would it be possible to also add a test demonstrating that a vrr user explicitly cannot see/execute the `Download file` button? This sort of does it implicitly, but it might be nice to have an explicit test as well.

hweng: @mcritchlow this is actually the old test unit forgot adding sign_in_curator, so I fixed here.
The vrr access only works on staging and production with token and user email. I'm not sure if it is possible to write the test.  I'm gonna dig into it.

mcritchlow: Sounds good!

app/views/dams_objects/_admin_download.html.erb
hweng: Do you mean to just add this one line conditional statement to helper method? Or do you mean to refactor the old code to the helper method? If it it the latter, i think it's better to create another refactoring ticket so that it will not block this new release.

hweng: the `vrr_user?` sounds good.

spec/features/dams_object_spec.rb
```diff
@@ -1717,3 +1719,47 @@
     expect(page).not_to have_content("Update DOI record")
   end
 end
+
+describe "vrr user who has authorized works should not see Download File and Embed button" do
+  let!(:user) { create_auth_link_user }
+
+  before(:all) do
+    @unit = DamsUnit.create pid: 'xx48484848', name: "Test Unit", description: "Test Description",
```
mcritchlow: I think this also needs to be deleted in the `after(:all)` block?

hweng: Updated, thanks for catching it.

hweng:
@mcritchlow Thanks for reviewing, the @unit was already deleted:
https://github.com/ucsdlib/damspas/pull/666/commits/2b7acf0da6a55bffad81cdb27facbe1849a11c57

VivianChu:
👍

PR #670, opened 2019-07-02
vrr: use rails logger for rest-client

mcritchlow:
It appears there are issues, yet to be determined, with batch processing vrr expirations.

This patch attaches the [rest-client gem](https://github.com/rest-client/rest-client#logging) to the Rails logger defined for the application. 

This will result in logging when API requests are made, hopefully giving us more insight into what may be going wrong.

Example output:

```
RestClient.get "no_aeon_api_provided/Requests", "Accept"=>"*/*; q=0.5, application/xml", "Accept-Encoding"=>"gzip, deflate", "X-AEON-API-KEY"=>"no_aeon_key_provided", "content-type"=>"text/json"
# => 200 OK |  0 bytes
=> ""
```

#### Local Checklist
- [ ] Code style checked?
- [x] QA-ed locally?

### DIFF
--- a/config/initializers/rest_client_logger.rb
+++ b/config/initializers/rest_client_logger.rb
@@ -1,0 +1,2 @@
+# Teach RestClient instances to use the configured Rails logger
+RestClient.log = Rails.logger


hweng:
Does all the http requests be automatically added to logging output? Or you need to call the method somewhere in the code to get output?

mcritchlow:
@hweng - great question. they get automatically logged. An alternative is to pass in a logger when a rest client is created: `resource = RestClient::Resource.new 'http://example.com/resource', log: Logger.new(STDOUT)`

But this initializer seemed like a nice way of:
- leaving the existing code alone
- ensuring that any additional rest-client usage/instances doesn't accidentally forget to add logging

see: https://github.com/rest-client/rest-client#logging

hweng:
@mcritchlow It sounds promising. Will see how the logger output looks like and how it is helpful. thank you for the explanation.

mcritchlow:
@hweng - yeah, i think it's at least going to hopefully be a helpful starting point. To your earlier point, we may also need to do more specific logging around the vrr issues, but this at least gets us whatever logging we can get from the rest client gem directly. Thanks!

Issue #634, opened 2019-05-16
Re-ordering of Role types in the UI


gamontoya:
From dams-metadata created by [hjsyoo](https://github.com/hjsyoo): ucsdlib/dams-metadata#55  

### Summary
Propose a re-ordering of Role types. The current D5 data dictionary is listed here: https://docs.google.com/spreadsheets/d/11B4lwD6qOHHvl3xmuU6aPTN_UMu-YMqBZ4XzwFDAKNE/edit#gid=1415126809.
The proposed new order for roles is listed in https://docs.google.com/spreadsheets/d/1sqdKcivHu36WnQweP-cslpsbpD26bwC1znvX52cQSHU/edit#gid=0.

### Rationale
Currently, Role types are ordered alphabetically for the most part, with an exception for Principal Investigator and Co-Principal Investigator. Like these exceptions, there are other roles that should appear higher in rank.

### Proposal
When these Roles are assigned for collections or objects, rank them in the UI in the following order:

- Creator
- Director
- Project Director
- Lead
- Author
- Principal Investigator
- Co-Principal Investigator
- Research Team Head
- Field Director
- Laboratory Director
- Researcher
- Research Team Member
- Analyst
- Data Contributor
- Data Manager
- Field Assistant
- Laboratory Assistant
- Photographer
- Advisor
- Thesis Advisor
- Contributor
- [All other Roles in alphabetical order]
- Other

gamontoya:
@hjsyoo You need to meet with Cristela and Matt P. before we'll take action on this request.

mcritchlow:
@hjsyoo - do you have any good example records that demonstrate some, or even most, of these? I'm going to get started on this, and mostly want to make sure we're checking metadata correctly (case, spelling, etc.). Thanks!

hjsyoo:
@mcritchlow Do you mean examples where the current order doesn't work well?
https://library.ucsd.edu/dc/collection/bb22212199
https://library.ucsd.edu/dc/collection/bb96275693
Or the official spellings of Roles? I was working off of this sheet, although we should get confirmation from @arwenhutt on whether this is the most authoritative version:
https://docs.google.com/spreadsheets/d/11B4lwD6qOHHvl3xmuU6aPTN_UMu-YMqBZ4XzwFDAKNE/edit#gid=1415126809

arwenhutt:
[...]

PR #672, opened 2019-07-03
Fix regression in role ordering

mcritchlow:
The previous implementation was not deleting roles that start with a
capital letter, which was an omission on my part. I've now inlined each condition and explicitly display the partial and delete the relationship as it is found (lowercase, capitalized).

I will also separately raise the issue of why these role values are not normalized. Ideally we shouldn't have to check for capitalization inconsistency.

Related: #634

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?


mcritchlow:
@hweng - I've added a test against the duplicate role name regression. Thanks again for the prod to do that :)

VivianChu:
:

hweng:
LGTM 👍

Issue #643, opened 2019-05-30
Display a Date Record Last Updated at bottom of object pages


hjsyoo:
### Descriptive summary

Display a Date Record Last Updated on bottom of all object landing pages. Frequently, data or metadata is updated in an object, but the Date:Issued and Publication year are intentionally not changed. Date Last Updated will succinctly indicate to the user when something on the page has been revised.

-------------

- Can make it a label. @hjsyoo will come up with display spec.
- SC&A is okay with this displaying on their content too.

hjsyoo:
@hjsyoo Post examples here

mcritchlow:
@hjsyoo @gamontoya - i think i'm going to start on this soon. I'll look at the internals for getting the date itself, however it would definitely help to have some kind of visual reference for what you're hoping to see when this is in place. Also, to clarify, is this _just_ for object pages, not collection pages?

hjsyoo:
@mcritchlow Yes, this is just for object pages. On coll pages, we were concerned that a user would have a hard time distinguishing whether the date info refers to updates on the page itself, or to anywhere in the entire collection.
I'm not sure what would work best visually - others might have a better sense for this. One minor consensus seemed to be that it should appear in a different style than the object metadata, and at the bottom of the page somewhere. I looked around briefly for examples on other sites, but had trouble finding any. @gamontoya Do you know of any places I should look?

mcritchlow:
Here's an update on my end. In our Solr records for objects, there are a few modified dates available, however they all seem to relate to the Solr document itself, rather than the actual object in the triplestore. Usually, this is _probably_ a safe date to rely on, but it won't always be. A triggered reindex for a collection that results in a new solr document being created for an object would mean a new 'updated' date, even if the record metadata wasn't actually updated.

So, at the moment, there's no quick win here unless we're OK with the 'mostly correct' modified date in the Solr document.

We do also have events being stored for objects in the triplestore, but as I'm starting to look into those a bit more, and discussing the events implementation with @lsitu (many thanks), it is clear that's not a easy win either unfortunately. It seems like historically we don't have a lot of event data, and we don't enforce tracking it consistently and safely, so at the moment it's not something we can depend on.

The Solr modified date might be a good starting point to get something out the door. A proper modified date that's provided by the triplestore appears to be a project unto itself.

hjsyoo:
[...]

PR #674, opened 2019-07-05
Revert custom RCDP collection sort

mcritchlow:
- The system modified date is not acceptable
- Leave tests pending so we can re-use them when we have a better
  solution

References #644 

#### Local Checklist
- [x] Code style checked?
- [ ] QA-ed locally?

#### What does this PR do?
This revert the change to use the `system_modified_dtsi` field in Solr for sorting RDCP collections. 

It does leave in place the tests that were written, it's just marked them as pending. Hopefully we can enable and reuse these tests once we have a different solution in place.

References #644 

A more appropriate solution is being discussed in #643 and a new PR will be made once we have that sorted out.


VivianChu:
👍

PR #678, opened 2019-07-10
gems: add pry-byebug

mcritchlow:
While looking into a few recent tickets I found that we didn't have this gem installed. I prefer the experience of using pry, but it's nice to have byebug integrated for debugging.


VivianChu:
👍

PR #683, opened 2019-07-12
Add VRR docs and cleanup README

mcritchlow:
#### Local Checklist
- [ ] Code style checked?
- [ ] QA-ed locally?
- [ ] Rebased with `master` branch?

#### What does this PR do?
- Create a `doc/` folder that for now holds the VRR documentation
- Add VRR blurb to README
- Remove old install instructions from README
- Clarify Docker instructions in README
- Replace old badges w/ current ones


README.md
```diff
@@ -1,213 +1,23 @@
...
-
-  Then navigate to localhost:3000
-
-  
-# Setup Instructions 
-
-## Windows
-_Let's not go there._
-
-## Apple/OSX
```
lsitu: Do we want to keep this as reference for setting up local development environment? But it's out dated and we may still need to clean it up bit if we want to keep it though. Thanks.

mcritchlow: @lsitu - it's a very good question, I'd be interested to hear what others think as well.

In my opinion, it's both outdated and no longer relevant. I think the `docker-compose` setup we have is far preferable than setting up each of the pieces separately, and much of the instructions at the bottom were more deployment/production related, which are documented better elsewhere by/for OPS. But that is just my opinion, if others on the team found value in those docs, perhaps we can work on updating them instead.

lsitu: @mcritchlow It sounds good if we have this kind of deployment/production related setup documented somewhere and link to it from README.

mcritchlow:
@lsitu - oh i appreciate that, but this is no rush. we can wait until @hweng and @VivianChu weigh in. This doesn't block VRR deployment. Honestly no one asked me to do this.. I just figured it would be good for this documentation to live in the repo instead of slack where it would otherwise be lost..

hweng:
LGTM 👍

PR #689, opened 2019-07-16
Title sorting ignores quotes and case

### LINKED ISSUE #661: Ignore characters when sorting by default title
### Descriptive summary

Currently in the default sort (by title), titles with apostrophes or quotes get sorted first. Is it possible to ignore initial characters and force the alpha sort?

### Example

sort

mcritchlow:
Fixes #661 

[MERGED](https://github.com/ucsdlib/dams4-solr/pull/5) ~~NOTE: There will need to be an accompanying PR against `dams4-solr` with these `schema.xml` changes. I haven't put that up yet, because I'd like feedback from the team on this approach before doing so. This will require a full reindex, so we want to get this right the first time, ideally..~~

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
- creates an `alphaStrippedSort` `fieldtype` which ignores double quotes/single quotes, and parens (for now), and filters case-sensitivity
- explicitly declares `title_ssi` as a field of type `alphaStrippedSort` for use in
  sorting instead of the default tokenizer/filters applied when analyzing `title_ssi`
- adds a test verifying the results illustrated in the attached screenshot

References #661 

#### Screenshots
pic-selected-190716-1301-44

### DIFF
--- a/solr_config/config/schema.xml
+++ b/solr_config/config/schema.xml
@@ -3,14 +3,14 @@
   <!-- NOTE: various comments and unused configuration possibilities have been purged
      from this file.  Please refer to http://wiki.apache.org/solr/SchemaXml,
      as well as the default schema file included with Solr -->
-  
+
   <uniqueKey>id</uniqueKey>
-  
+
   <fields>
     <field name="id" type="string" stored="true" indexed="true" multiValued="false" required="true"/>
     <field name="id_t" type="string" stored="true" indexed="true" multiValued="false" required="true"/>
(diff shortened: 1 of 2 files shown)


mcritchlow:
@gamontoya - does the screenshot above look like what you'd expect for sorting?

gamontoya:
@mcritchlow Yes, right on!

hweng:
:+1:

PR #692, opened 2019-07-22
Fixes #642 - Create conditional copyright boilerplate for RDCP.

### LINKED ISSUE #642: Create conditional copyright boilerplate, for objects with CC license
### Descriptive summary

Change display of "Copyright Constraint(s) on Use" boilerplate for Research Data Collections, wherever Rights holder is UC Regents and Creative Commons license is applied, to

"This work is protected by U.S. Copyright Law (Title 17, U.S.C.), and made available under a Creative Commons (CC) license. Use of this work beyond that allowed by "fair use" or the CC license requires written permission of UC San Diego. Responsibility for obtaining permissions and any use and distribution of this work rests exclusively with the user and not the UC San Diego Library. Inquiries can be made to the UC San Diego Library program having custody of the work." 

Where Rights holder is an individual(s) and CC is applied, use the variant:

"This work is protected by U.S. Copyright Law (Title 17, U.S.C.), and made available under a Creative Commons (CC) license. Use of this work beyond that allowed by "fair use" or the CC license requires written permission of the copyright holder(s). Responsibility for obtaining permissions and any use and distribution of this work rests exclusively with the user and not the UC San Diego Library. Inquiries can be made to the UC San Diego Library program having custody of the work."

### Rationale

The current text is confusing wherever CC has been applied. The revision above was suggested by Katie Fortney, UC Copyright Officer, to clarify the text for users. Aside from references to the Creative Commons license, the new text states that permission should be requested from UC San Diego. The current text for the UC authority lists "UC Regents". "UC San Diego" is almost equally vague, but it indicates that the authority is local. Library staff responding to inquiries should be aware that the current delegated authority is the UCSD Office of Innovation and Commercialization, formerly known as Tech Transfer.

-----------

- Effort is low for now unless conditional is more complicated.

hweng:
Fixes #642

#### Local Checklist
- [x] Code style checked?
- [x] QA-ed locally?

#### What does this PR do?
Create conditional copyright boilerplate, for objects with CC license for RDCP collection only.

References #642 

### DIFF
--- a/app/views/shared/fields/_copyright.html.erb
+++ b/app/views/shared/fields/_copyright.html.erb
@@ -10,6 +10,8 @@
 	 end
 
 	 fieldData = @document["#{prefix}copyright_tesim"]
+	 rights_holder = @document["#{prefix}rightsHolder_tesim"]
+	 unit_code = @document["#{prefix}unit_code_tesim"]
 
 	 if fieldData != nil
 
@@ -24,7 +26,16 @@
 			<p><strong><%= copyright['status'].gsub(/--/, ' – ') %></strong></p>
 		<% end %>
 		<p><%= copyright['purposeNote'] %></p>
-		<p><%= copyright['note'] %></p>
+    
+    <% if unit_code == "rdcp" %>
+			<% if rights_holder && rights_holder == "UC Regents"%>
+			  <p>This work is protected by U.S. Copyright Law (Title 17, U.S.C.), and made available under a Creative Commons (CC) license. Use of this work beyond that allowed by "fair use" or the CC license requires written permission of UC San Diego. Responsibility for obtaining permissions and any use and distribution of this work rests exclusively with the user and not the UC San Diego Library. Inquiries can be made to the UC San Diego Library program having custody of the work.</p>
+			<% else %>
+			  <p>This work is protected by U.S. Copyright Law (Title 17, U.S.C.), and made available under a Creative Commons (CC) license. Use of this work beyond that allowed by "fair use" or the CC license requires written permission of the copyright holder(s). Responsibility for obtaining permissions and any use and distribution of this work rests exclusively with the user and not the UC San Diego Library. Inquiries can be made to the UC San Diego Library program having custody of the work.</p>
+			 <% end %>
+		<% else %>
+		  <p><%= copyright['note'] %></p>
+		<% end %>
 	<%
 
 		 concat htmlClose.html_safe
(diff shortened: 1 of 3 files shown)


app/views/shared/fields/_copyright.html.erb
```diff
@@ -24,7 +26,16 @@
 			<p><strong><%= copyright['status'].gsub(/--/, ' – ') %></strong></p>
 		<% end %>
 		<p><%= copyright['purposeNote'] %></p>
-		<p><%= copyright['note'] %></p>
+    
+    <% if unit_code == "rdcp" %>
```
mcritchlow: I think a test demonstrating this change _might_ be helpful, primarily in the future when we need to migrate to a new application. Basically just something demonstrating that we're doing something unique with respect to RDCP copyright. Do others have an opinion on this?

app/views/shared/fields/_copyright.html.erb
```diff
@@ -24,7 +26,16 @@
...
+			<% if rights_holder && rights_holder == "UC Regents"%>
+			  <p>This work is protected by U.S. Copyright Law (Title 17, U.S.C.), and made available under a Creative Commons (CC) license. Use of this work beyond that allowed by "fair use" or the CC license requires written permission of UC San Diego. Responsibility for obtaining permissions and any use and distribution of this work rests exclusively with the user and not the UC San Diego Library. Inquiries can be made to the UC San Diego Library program having custody of the work.</p>
```
mcritchlow: As far as I can tell, the difference in text here, is `UC Regents` gets a `...written permission of UC San Diego` string as opposed to `...written permission of the copyright holder(s)` when the rights holder is anything else (but still an RDCP record).

Is it worth considering pulling this out into a variable, or even a small function?

hweng: Thanks for the suggestion, I've updated it to put the differential into variable.

app/views/shared/fields/_copyright.html.erb
hweng: Thanks for the suggestion. It's not hard to write the test for it. But I feel this is not like VRR updates, it's just a trivial view changes in the .erb file. It will be too much tests in our codebase if add such test suit case. What do you think?

mcritchlow: Yeah, that's why I said might. It feels on the fence. The reason I lean towards a test is this change is, in a sense, business logic. Albeit fairly trivial business logic :) 

My real concerns are:
- losing track of stuff like this in the future. But it is so small, at this point, that I think your pushback is fair. 
- there being more "If RDCP then do this, else do that" stuff coming in the future. My current branch waiting on Solr reindexing, for example, does this for sorting. 

So like I said, I don't have a strong opinion we need a test or two for this, but those are the reasons I mention it. @lsitu and @VivianChu , what are your thoughts?

VivianChu: I don't have a strong opinion for a test either.  However, I think it would be nice to have a test since the change is business logic.

VivianChu:
👍

